PR #155, opened 2026-04-23
fix(config): swap chat model to moonshotai/kimi-k2.6

coleam00:
## Linked issue

Fixes #153

## Summary

Swaps the pinned chat completion model from `anthropic/claude-sonnet-4.6` to `moonshotai/kimi-k2.6` across all configuration and documentation. This is a governance-authorized model change approved by Cole (owner) to improve tool-use performance and reduce cost.

## How this solves the issue

Issue #153 requested updating the hardcoded `CHAT_MODEL` constant in `config.py` and ensuring all documentation references (`CLAUDE.md`, factory validator commands) align with the new model. The diff changes exactly 5 files:

1. `app/backend/config.py:70` — `CHAT_MODEL` constant updated from `anthropic/claude-sonnet-4.6` to `moonshotai/kimi-k2.6`
2. `app/backend/llm/openrouter.py` — module docstring updated to name the new model
3. `CLAUDE.md` — RAG Pipeline Invariant #4 updated to pin Kimi K2.6
4. `.archon/commands/dark-factory-validate.md` — validator invariant check updated so it no longer rejects Kimi K2.6
5. `.archon/commands/dark-factory-fix-pr-issues.md` — hard-invariant check updated in lockstep

All grep checks confirm zero `claude-sonnet` matches remain in `app/`, `CLAUDE.md`, or `.archon/commands/`.

## Test plan

- [x] `grep -rn 'claude-sonnet' app/ CLAUDE.md .archon/commands/` → 0 matches
- [x] `grep -rn 'kimi-k2' app/ CLAUDE.md .archon/commands/` → 5 matches in the 5 expected files
- [x] `uv run ruff check .` — 0 issues (backend, from `app/backend/`)
- [x] `uv run ruff format --check .` — pass (1 unrelated blank-line fix in test file)
- [x] `uv run mypy .` — Success: no issues found in 69 source files
- [x] `uv run pytest tests -xvs` — 218 passed, 61 skipped, 0 failed
- [x] `bun run tsc --noEmit` — 0 TypeScript errors
- [x] `bun x biome check src` — 38 files checked, no errors
- [x] `bun run test` — 98 tests passed across 12 test files

## Agent-browser regression

- [ ] Full happy-path regression passes (sign-in → new conversation → ask question → streaming response → citations → citation modal with embedded player)

## Dependencies

No new dependencies added or removed.

## Governance / protected files

[...]

PR #163, opened 2026-04-24
fix(sse): emit keepalive heartbeats during tool-call streaming and tool execution

coleam00:
## Linked issue

Blocks the production regression caller reported after the Kimi K2.6 swap (#155): "agent thinks briefly, loading indicator goes away, no tokens stream." Not a separate GitHub issue yet - filing this directly as a hotfix PR.

## Summary

`stream_chat` in `backend/llm/openrouter.py` only yields `data:` chunks when the model emits content tokens. Kimi K2.6 frequently streams tool_call args silently for tens of seconds per round, then awaits the executor (embedding + Postgres query), then does it again. On tool-heavy queries ("How does Cole recommend building AI agents?") the backend can be actively processing for 60-140+ seconds while emitting zero SSE bytes. Browsers and reverse proxies idle-timeout the socket, the frontend marks the stream as done, the loading indicator disappears, and the user sees an empty assistant bubble.

This PR adds standard SSE comment heartbeats (`: keepalive\n\n`) during the silent phases so the socket stays warm until real tokens arrive.

## Reproduction captured from prod

- Via `curl --max-time 240 --no-buffer` (no idle timeout): first content token at t=140s, sources at t=158s, done at t=158s. Works.
- Via agent-browser (standard fetch + ReadableStream): body length stuck at 1064 chars (only the user-bubble) for 2+ minutes, then `os error 10060` (connection dropped).

## How this solves the issue

Two heartbeats, both emitting the same spec-valid SSE comment `": keepalive\n\n"`:

1. **During tool_call arg streaming** - throttled to a 5 second cadence so a chatty tool_call round doesn't flood the stream. Emitted inside the `async for chunk in stream:` loop when only `delta.tool_calls` is present (no content).
2. **Before each `await tool_executor(...)`** - covers the embedding + DB-query window where the coroutine is suspended. One heartbeat per tool call.

The frontend parser (`useStreamingResponse.ts`) only matches `startsWith('event:')` or `startsWith('data:')`; lines starting with `:` are silently skipped, so comments never render anywhere. `_extract_text_from_sse` in `routes/messages.py` also ignores non-`data:` chunks, so keepalives never leak into the persisted assistant text.

## Test plan

5 new tests in `test_sse_heartbeat.py`:

- `test_keepalive_emitted_while_model_streams_tool_call_args` - drives a 12-fragment tool_call round with 1 second delays; asserts a keepalive arrives before the first content token
- `test_keepalive_emitted_before_tool_execution` - asserts a keepalive precedes the executor await
- `test_keepalive_does_not_replace_content_tokens` - sanity: content reconstructs to "Answer here." from data chunks; `[DONE]` still terminates
- `test_keepalive_format_is_spec_valid_sse_comment` - locks in the exact `": keepalive\n\n"` format
- `test_extract_text_skips_keepalive_comments` - the persist path skips keepalives

- [x] `uv run pytest tests` - **298 passed**, 61 skipped, 0 failed
- [x] `uv run ruff check .` - clean
- [x] `uv run ruff format --check .` - clean
- [x] `uv run mypy .` - no issues in 77 source files

## Agent-browser regression

- [ ] Post-merge: rerun "How does Cole recommend building AI agents?" in a browser. Expect first keepalive within ~5 seconds, connection stays open through the 2+ minute tool-call phase, final answer renders with Sources chip.
- [ ] Post-merge: rerun a fast-first-token query ("What is RAG?") to confirm no regression on the happy path where content streams immediately.

## Dependencies

None. Only uses `time.monotonic()` from the standard library.

## Governance / protected files

- [x] No protected files modified (only `llm/openrouter.py` and a new test file)
- [x] PR size within 500 lines (2 files, 261 insertions, 0 deletions)
- [x] No weakening of authentication, authorization, or the 25 msg/day rate limit
- [x] No provider or model change; stream shape unchanged from the client's perspective (comments are ignored by the parser)

### DIFF
--- a/app/backend/llm/openrouter.py
+++ b/app/backend/llm/openrouter.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import json
 import logging
+import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from typing import Any, cast
 
@@ -130,6 +131,18 @@ async def stream_chat(
 
     tool_calls_made = 0
     tokens_yielded = 0
+    # Heartbeat cadence: Kimi K2.6 regularly goes 60-140s of silent tool-call
+    # streaming + tool execution before emitting the first user-visible text
+    # token. Browsers and reverse proxies idle-timeout SSE connections after
+    # ~60s of no bytes. Emitting SSE comment lines (`: <text>\n\n`) every few
+    # seconds keeps the socket warm. Comments are spec-valid SSE that clients
+    # ignore, so the frontend is unaffected.
+    HEARTBEAT_INTERVAL_SECONDS = 5.0
+    last_heartbeat_at = time.monotonic()
+
+    def _heartbeat_due() -> bool:
+        return (time.monotonic() - last_heartbeat_at) >= HEARTBEAT_INTERVAL_SECONDS
+
     try:
         while True:
             # Once the per-turn cap has been reached, stop declaring tools on
@@ -156,6 +169,7 @@ async def stream_chat(
                     assistant_text_parts.append(delta.content)
                     tokens_yielded += 1
                     yield f"data: {json.dumps(delta.content)}\n\n"
+                    last_heartbeat_at = time.monotonic()
                 if delta and delta.tool_calls:
                     for tc in delta.tool_calls:
                         slot = pending.setdefault(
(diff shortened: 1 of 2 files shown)


coleam00:
## Archon PR Validation Report

**Verdict**: APPROVE

### Summary
SSE keepalive heartbeats correctly address both silent windows in `stream_chat` — tool-call argument streaming and tool execution — that caused browser/proxy idle timeouts on Kimi K2.6 tool-heavy queries. The fix is 27 lines of production code using only `time.monotonic()`, emits spec-valid SSE comments that all consumers already ignore, and ships with 5 regression tests (all passing). No issues found, no regressions.

### Bug Confirmation
| Claim | Main | Feature |
|-------|------|---------|
| Zero SSE bytes during tool_call arg streaming | BUG CONFIRMED | FIXED (throttled 5s keepalive) |
| Zero SSE bytes during tool execution await | BUG CONFIRMED | FIXED (unconditional keepalive before executor) |
| Frontend/persist path ignore SSE comments | CONFIRMED | CONFIRMED (test + code review) |

### Issues
No blocking issues found.

### Notes
- Browser E2E was blocked on both branches (no local Postgres available). Verification via code analysis + unit test suite provides high confidence given the deterministic nature of this bug.
- Post-merge agent-browser regression recommended per PR checklist.

---
_Validated by archon-validate-pr workflow_

PR #226, opened 2026-05-20

### LINKED ISSUE #225: [smoke-test] invalidate_cache() should log entry count cleared
## Summary

The `invalidate_cache()` helper in `app/backend/rag/retriever_hybrid.py` clears the module-level `_video_cache` dict and logs an info message. The log message doesn't include how many entries were cleared, which makes it harder to verify the cache was actually populated before invalidation when debugging.

## Expected

Update `invalidate_cache()` so its log message includes the count of entries cleared.

## Acceptance

- The log message after invalidation includes the integer count of entries removed (e.g. `'Hybrid retriever video cache invalidated (3 entries cleared).'`)
- No behavior change beyond the log message (still clears the same dict, still uses `logger.info`)
- Existing tests (if any) still pass

## Out of scope

- Don't refactor `_video_cache` or change its data shape
- Don't add new tests just for this — a one-line log change doesn't warrant a new test file
- Don't touch any other module

## Why this is a smoke-test issue

This issue is being filed as the smoke test for a new mixed-provider benchmark workflow. It's intentionally tiny and tractable so the workflow wiring can be validated end-to-end without committing the experiment's real budget. Workflow dispatch is via `benchmark-KK` (Kimi/Kimi cell — cheapest cell to validate first).

coleam00:
Fixes #225

**Source run-id:** `b151e664-bd7d-43a8-a47a-a3b505d38e49`

(Backfilled 2026-05-20 to validate the evaluator wiring fix)

### DIFF
--- a/app/backend/rag/retriever_hybrid.py
+++ b/app/backend/rag/retriever_hybrid.py
@@ -31,8 +31,9 @@ _video_cache: dict[str, dict[str, str]] = {}
 def invalidate_cache() -> None:
     """Clear the video metadata cache."""
     global _video_cache
+    count = len(_video_cache)
     _video_cache.clear()
-    logger.info("Hybrid retriever video cache invalidated.")
+    logger.info(f"Hybrid retriever video cache invalidated ({count} entries cleared).")
 
 
 async def retrieve_hybrid(


coleam00:

Textbook execution of a smoke-test issue. The change is exactly two lines: snapshot len(_video_cache) into a local before .clear(), then interpolate it into the existing logger.info message — matching the acceptance example string almost verbatim. Order of operations is correct (count captured pre-clear), scope is surgical (one file, one function), and the author appropriately declined to add tests as the issue requested.

Weaknesses are minor and largely stylistic: f-string in a logger call eagerly builds the message regardless of log level (lazy '%s' formatting would be marginally more idiomatic for logging), and plan/impl fidelity isn't assessable because there's no separate plan commit in the history. For a Kimi/Kimi cell validation of a one-line log change, this is about as clean as it gets.

## Scores

|---|---|---|
| solves_issue | 10/10 | Log message now includes the integer count of entries cleared, matching the acceptance criteria example almost verbatim ('Hybrid retriever video cache invalidated (3 entries cleared).'). |
| root_cause | 10/10 | Issue is intrinsically a log-message change; the PR captures count via len() before clearing and emits it in the same logger.info call. There is no symptom-vs-root-cause ambiguity here. |
| scope_discipline | 10/10 | Single file touched (retriever_hybrid.py), +2/-1 lines, exactly the function called out. No drive-by edits. |
| subtle_correctness | 9/10 | Correctly snapshots len(_video_cache) BEFORE .clear() — reversing those lines would have been a silent bug logging 0. f-string interpolation in logger.info is fine for an info-level dev message; no async/typing concerns. |
| code_quality | 8/10 | Readable and idiomatic. Minor nit: using an f-string in logger.info instead of '%s' lazy formatting is a small style miss (eager string build even when log level is filtered), but it matches the rest of the file's tone. |
| test_discipline | 8/10 | Issue explicitly says 'Don't add new tests just for this — a one-line log change doesn't warrant a new test file.' PR correctly followed that scoping instruction; no test gaming observed. Penalty avoided because adding tests would have violated scope. |
| plan_impl_fidelity | 5/10 | Not visible — commit log shows only a single 'impl: address issue' commit with no separate plan commit to compare against, so plan↔impl alignment cannot be assessed. |

---

coleam00:

This is an excellent, minimal execution of a trivial smoke-test issue. The two-line diff exactly matches both the issue's acceptance example and the plan's prescribed f-string, with the count captured before clear() so the logged value correctly reflects pre-invalidation size. Scope discipline is perfect — one file, one function, no tangential edits — and the no-tests choice correctly honors the issue's explicit guidance.

The only minor weakness is the use of an f-string rather than lazy %-style logging (logger.info('... (%d entries cleared).', count)), which would skip string formatting when the log level is filtered out. That's a stylistic nit on a debug-grade info log and doesn't materially affect correctness. As a smoke test for the KK (Kimi/Kimi) cell, the workflow produced a clean, faithful implementation.

## Scores

|---|---|---|
| solves_issue | 10/10 | The fix produces exactly the log message format shown in the issue's acceptance example, including the integer count and the 'entries cleared' phrasing. Behavior is otherwise unchanged. |
| root_cause | 10/10 | The issue is itself a log-message change; capturing len() before clear() and embedding it in the same logger.info call is the structurally correct fix with no symptom-papering. |
| scope_discipline | 10/10 | Single file, two-line diff (+2/-1), touches only invalidate_cache() as required. No tangential edits, no out-of-scope refactors. |
| subtle_correctness | 9/10 | Count is captured before clear() so it reflects the pre-invalidation size correctly. f-string is safe here (no untrusted input). Minor nit: using lazy %-style logging is sometimes preferred, but f-string is fine and idiomatic in this codebase. |
| code_quality | 9/10 | Readable, idiomatic, matches surrounding style. Uses f-string consistent with modern Python conventions. The redundant `global _video_cache` declaration was preserved (existing code), which is appropriate given scope discipline. |
| test_discipline | 8/10 | No tests added, which the issue explicitly forbids ('Don't add new tests just for this'). Existing tests are untouched and not gamed. Correctly followed the issue's test guidance. |
| plan_impl_fidelity | 10/10 | Diff matches the plan verbatim: count = len(_video_cache) inserted before clear(), logger.info updated to the exact f-string suggested in the plan. No drift between plan and impl. |

---

PR #230, opened 2026-05-21

### LINKED ISSUE #227: fix(rag): _video_cache has a thundering-herd race + unbounded growth
## Summary

## Current behavior
- No eviction / maximum size.

## Acceptance

## Out of scope
- Don't add a new module.

coleam00:
Fixes #227

**Source run-id:** `133fe045-4098-4074-a403-043e3409bf76`

### DIFF
--- a/app/backend/rag/retriever_hybrid.py
+++ b/app/backend/rag/retriever_hybrid.py
@@ -16,8 +16,9 @@ cosine fallback).
 
 from __future__ import annotations
 
+import asyncio
 import logging
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 
 from backend.config import HYBRID_K_CONSTANT, HYBRID_OVERFETCH_FACTOR, KEYWORD_LANGUAGE
 from backend.db import repository
@@ -25,7 +26,9 @@ from backend.db import repository
 logger = logging.getLogger(__name__)
 
 # Module-level video metadata cache (populated on demand per chunk result)
-_video_cache: dict[str, dict[str, str]] = {}
+MAX_VIDEO_CACHE_SIZE = 256
+_video_cache: OrderedDict[str, dict[str, str]] = OrderedDict()
+_inflight: dict[str, asyncio.Future[dict[str, str]]] = {}
 
 
 def invalidate_cache() -> None:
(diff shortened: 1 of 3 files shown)


coleam00:

A high-quality, surgical fix that addresses both the thundering-herd race and the unbounded-growth bug at their root with the canonical patterns (asyncio.Future single-flight + OrderedDict LRU). The implementation closely tracks the plan, with one thoughtful deviation (wait path reads from the future, not the cache) that is documented and actually more robust than the plan as written. Tests are tightly targeted to the acceptance criteria and exercise the real concurrency path via asyncio.gather + a sleeping mock.

Weaknesses are minor: asyncio.CancelledError during the leader's await would leave waiters hanging (a BaseException edge case not covered by `except Exception`); and the LRU eviction test is mildly slow at 256+1 iterations but still acceptable. Overall this is one of the cleaner fixes I'd expect from a senior engineer — close to production-grade as-is.

## Scores

|---|---|---|
| solves_issue | 10/10 | All four acceptance criteria met: concurrent misses for the same video_id coalesce into one repository.get_video() call via an asyncio.Future single-flight gate, cache is bounded by MAX_VIDEO_CACHE_SIZE=256 with LRU eviction via OrderedDict.popitem(last=False), hits remain O(1) with no added awaits (just move_to_end), distinct video_ids are not serialized, and a regression test (test_single_flight_concurrent_misses) is added. |
| root_cause | 10/10 | Fixes both root causes structurally: the race condition is resolved with a per-key inflight Future map (canonical single-flight pattern), and unbounded growth is resolved with OrderedDict + size cap (canonical LRU). No band-aids — both invariants are enforced at the cache layer itself. |
| scope_discipline | 10/10 | Surgical: only the three files required by the fix were touched — retriever_hybrid.py for the core change, conftest.py to extend the existing cache-reset fixture for the new _inflight dict, and the test file for regressions. Schema of cached entries unchanged, no new module, no refactor of the result-construction loop. |
| subtle_correctness | 8/10 | Thoughtful: wait path reads the value directly from the awaited future (with an explicit comment) to avoid races with concurrent eviction or invalidate_cache() — a robustness improvement over the plan. try/except/finally correctly sets exception on the future, re-raises, and always cleans up _inflight. Minor gap: asyncio.CancelledError (BaseException, not caught by `except Exception`) during the leader's await would leave the future unset and any waiters hanging — but the finally still pops the gate. invalidate_cache() intentionally does not clear _inflight, which is documented and correct. |
| code_quality | 9/10 | Readable, well-commented at non-obvious decision points (hit/wait/leader paths labeled; the wait-path-reads-from-future comment explains a subtle invariant). Idiomatic Python (OrderedDict, asyncio.get_running_loop().create_future). No SQL, no fetch — fits codebase conventions. Module-level constant MAX_VIDEO_CACHE_SIZE is uppercase as expected. |
| test_discipline | 9/10 | Three targeted tests: single-flight concurrent miss (the explicit acceptance criterion, with asyncio.gather and a sleeping mock to force the race), LRU eviction at the cap boundary verifying v0 is evicted when a 257th entry lands, and a cache-hit test confirming repository.get_video is called exactly once across repeated queries. Tests mock at the right boundary (repository functions) and don't game the cache state — fixture properly resets both _video_cache and _inflight. |
| plan_impl_fidelity | 9/10 | Implementation faithfully executes the plan: OrderedDict swap, _inflight map, hit/wait/leader path structure, try/except/finally with set_result/set_exception/pop, conftest fixture extension, and the same test trio (single-flight, bounded-growth, hit-path). One intentional, documented deviation: the wait path uses `await _inflight[video_id]` to consume the future's value directly rather than re-reading the cache after the await — explicitly justified in a code comment as protection against concurrent eviction or invalidate_cache() during the wait. |

---

PR #231, opened 2026-05-21

### LINKED ISSUE #227: fix(rag): _video_cache has a thundering-herd race + unbounded growth

coleam00:
Fixes #227

**Source run-id:** `9eeae813-8665-4725-9d96-10efaaf9646b`

### DIFF
--- a/app/backend/rag/retriever_hybrid.py
+++ b/app/backend/rag/retriever_hybrid.py
@@ -16,23 +16,82 @@ cosine fallback).
 
 from __future__ import annotations
 
+import asyncio
 import logging
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 
 from backend.config import HYBRID_K_CONSTANT, HYBRID_OVERFETCH_FACTOR, KEYWORD_LANGUAGE
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
 
-# Module-level video metadata cache (populated on demand per chunk result)
-_video_cache: dict[str, dict[str, str]] = {}
+# Why: bound the cache so long-running processes don't grow it without limit.
+MAX_VIDEO_CACHE_SIZE = 1024
+
+# Module-level video metadata cache (populated on demand per chunk result).
+# OrderedDict gives O(1) LRU touch via move_to_end + popitem(last=False).
+_video_cache: OrderedDict[str, dict[str, str]] = OrderedDict()
+
+# Single-flight coordination: while a video_id's metadata is being fetched, the
+# first task stores a Future here; concurrent misses await that same Future
+# instead of issuing a duplicate repository.get_video() call.
+_inflight: dict[str, asyncio.Future[dict[str, str]]] = {}
 
 
 def invalidate_cache() -> None:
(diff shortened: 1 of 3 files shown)


coleam00:

An excellent execution of a well-scoped fix. The single-flight pattern (per-key Future stored before the await, with exception propagation via set_exception and finally-pop for retryability) is exactly the right primitive, and the OrderedDict-based LRU is the simplest std-lib solution that satisfies the bounded-size requirement. Scope is surgical, tests are targeted and non-trivial (a gated AsyncMock to force concurrent misses is the right setup), and plan-fidelity is near-perfect down to the helper signature and test names.

Minor weaknesses are cosmetic: the helper's warning log dropped the chunk_id (a deliberate trade-off acknowledged in the plan), and the final test asserts `'v1' not in _inflight` even though the fixture uses _CHUNK_A whose video_id is 'v1' — accurate but coupled to fixture internals. Overall this is a strong reference implementation for the OO benchmark cell.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly addresses both acceptance criteria: single-flight coordination via per-key Futures eliminates the thundering herd, and an OrderedDict-based LRU with MAX_VIDEO_CACHE_SIZE=1024 bounds growth. Cache hits remain O(1) with no awaits, and distinct video_ids are not serialized. |
| root_cause | 9/10 | Structurally sound: per-key in-flight Futures stored before the await is the canonical asyncio single-flight pattern, and OrderedDict + popitem(last=False) is the textbook LRU. Not a band-aid — the race window is genuinely closed. |
| scope_discipline | 10/10 | Touches only the three files called out in the plan (retriever_hybrid.py, conftest.py, test_retriever_hybrid.py). Result-construction loop is preserved, no new module, no schema changes, no unrelated refactors. |
| subtle_correctness | 9/10 | Handles exception path correctly (set_exception + finally pop so retries work), uses asyncio.get_running_loop() not get_event_loop(), and reasons correctly about invalidate_cache leaving in-flight waiters intact. Minor nit: on exception the cache stays empty but the Future receives the exception — that's the intended retry behavior. No obvious foot-guns. |
| code_quality | 9/10 | Clear docstring explaining the three cache-state branches, sensible comments on the 'Why' for the cap and the invalidate semantics, idiomatic Python. Removed an unused `global` declaration as a small bonus. Reads as production-grade. |
| test_discipline | 10/10 | Three targeted regression tests covering the exact behaviors the issue calls out: single-flight (mock count == 1 under 5 concurrent callers using an asyncio.Event gate), LRU eviction at a monkeypatched bound, and the failed-fetch retry path. Uses the autouse reset fixture for isolation. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan essentially verbatim: same module-level constants and types, same _get_video_meta helper structure with try/except/finally, same three tests with the exact strategies described (Event gate, monkeypatched bound, flaky get_video). Even the decision to log without chunk_id inside the helper follows the plan's recommendation. |

---

PR #232, opened 2026-05-21

### LINKED ISSUE #227: fix(rag): _video_cache has a thundering-herd race + unbounded growth

coleam00:
Fixes #227

**Source run-id:** `443f8b23-fbb5-4a79-a456-a37cddb27ec0`

### DIFF
--- a/app/backend/rag/retriever_hybrid.py
+++ b/app/backend/rag/retriever_hybrid.py
@@ -16,22 +16,90 @@ cosine fallback).
 
 from __future__ import annotations
 
+import asyncio
 import logging
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
+from collections.abc import AsyncIterator, Callable
+from contextlib import asynccontextmanager
 
 from backend.config import HYBRID_K_CONSTANT, HYBRID_OVERFETCH_FACTOR, KEYWORD_LANGUAGE
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
 
-# Module-level video metadata cache (populated on demand per chunk result)
-_video_cache: dict[str, dict[str, str]] = {}
+# Maximum number of video metadata entries to retain in the LRU cache.
+MAX_CACHE_SIZE = 256
+
+
+class BoundedVideoCache:
+    """LRU-bounded video metadata cache with per-key async single-flight.
+
+    Cache hits are O(1) lock-free dict lookups. On miss, concurrent callers
+    for the same ``video_id`` share a single downstream fetch via a per-key
+    ``asyncio.Lock`` — distinct ``video_id`` keys never serialize against
+    each other.
+    """
+
+    def __init__(self, maxsize: int) -> None:
+        self._maxsize = maxsize
+        self._data: OrderedDict[str, dict[str, str]] = OrderedDict()
+        self._locks: dict[str, asyncio.Lock] = {}
+
+    def get(self, video_id: str) -> dict[str, str] | None:
+        value = self._data.get(video_id)
+        if value is not None:
+            self._data.move_to_end(video_id)
+        return value
+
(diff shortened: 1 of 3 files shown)


coleam00:

Strong PR that hits the issue's acceptance criteria cleanly. The BoundedVideoCache abstraction is well-designed: OrderedDict-backed LRU with per-key asyncio.Lock single-flight, double-checked locking, and a clean asynccontextmanager API that keeps the hydration-loop diff small. Scope is surgical (3 files, no drift), code quality is production-grade with good docstrings explaining the why, and the regression test directly exercises the thundering-herd scenario via asyncio.gather + asyncio.sleep overlap. Implementation faithfully executes the plan. Weaknesses are minor: no test for LRU eviction or distinct-key parallelism, and a small deviation from the plan's try/finally pattern around the fetch (the impl's lock-release-then-retry approach is correct but not what the plan specified). Overall a high-quality fix that demonstrates good understanding of asyncio concurrency primitives.

## Scores

|---|---|---|
| solves_issue | 10/10 | All four acceptance criteria are satisfied: single-flight via per-key asyncio.Lock guarantees at most one repository.get_video() call for concurrent misses on the same video_id; an OrderedDict-backed LRU caps size at 256; cache hits are O(1) lock-free dict lookups with no added awaits; distinct video_ids hold distinct locks and never serialize; and a regression test asserts the single-flight behavior. |
| root_cause | 9/10 | Replaces the bare dict with a purpose-built BoundedVideoCache that addresses both root causes (race + unbounded growth) structurally rather than papering over them. The classic double-check + per-key lock pattern is the textbook single-flight solution; LRU eviction via OrderedDict.popitem(last=False) is the canonical bounded-cache approach. |
| scope_discipline | 10/10 | Touches exactly the three files needed: retriever_hybrid.py (the fix), conftest.py (fixture updated to reset both data + lock state), and the test file (regression test added). The hydration loop's result-construction is preserved as the issue demanded. No drive-by edits. |
| subtle_correctness | 8/10 | Handles the tricky bits well: double-check inside the lock to prevent post-release re-fetches; lock cleanup on success; on exception the per-key lock stays in _locks but the next caller retries (correct, just slightly leaky if the failure persists). One minor nit: the plan called for try/finally around the fetch to guarantee release, and the impl relies on the surrounding async-with for lock release instead — works correctly because the not-yet-released lock entry will be picked up by the next caller, but a try/except calling release(unknown_meta) would have been cleaner. |
| code_quality | 9/10 | Clean class design with clear docstrings explaining the release-fn contract, proper type annotations (AsyncIterator, Callable), idiomatic asynccontextmanager usage, and good naming (get_or_lock, invalidate). Comments explain the why (single-flight, lock-free fast path) rather than restating the what. |
| test_discipline | 8/10 | Regression test is well-targeted: uses asyncio.sleep(0.05) inside the mocked get_video to guarantee overlap, fires two concurrent retrieve_hybrid() calls via asyncio.gather, and asserts mock_video.await_count == 1 plus result correctness. Missing: no explicit test for LRU eviction (bounded-size behavior is asserted only by the type-system, not exercised), and no test confirming distinct video_ids don't serialize. |
| plan_impl_fidelity | 9/10 | Implementation matches the plan almost verbatim — BoundedVideoCache class, OrderedDict + per-key Lock map, MAX_CACHE_SIZE = 256, get() fast path, get_or_lock context manager yielding (cached, release_fn), invalidate() clearing both, conftest update, and the named single-flight regression test. Only deviation: plan called for try/finally around the fetch to guarantee release; impl relies on the surrounding async-with for lock release without a try/finally, which works but slightly drifts from the plan's described error-handling pattern. |

---

PR #233, opened 2026-05-21
[bench-OO] fix(ingest): clean up orphan video row on embedding failure in /ingest/from-url

### LINKED ISSUE #228: fix(ingest): /ingest/from-url leaves an orphan video row when embedding fails
## Summary

## Current behavior

## Acceptance

## Out of scope
- No other behavior changes.

coleam00:
Fixes #228

**Source run-id:** `105e2abc-2d70-4e73-87fb-c5363f611932`

### DIFF
--- a/app/backend/routes/ingest.py
+++ b/app/backend/routes/ingest.py
@@ -270,6 +270,8 @@ async def ingest_from_url(body: IngestFromUrlRequest) -> IngestFromUrlResponse:
         embeddings = embed_batch(chunk_texts)
     except Exception as exc:
         logger.error("Embedding batch failed for video '%s': %s", title, exc)
+        # Clean up the orphan video record to avoid leaving cruft
+        await repository.delete_video(video_id)
         raise HTTPException(
             status_code=502,
             detail=f"Embeddings API request failed: {exc}",
--- a/app/backend/tests/test_video_ingest.py
+++ b/app/backend/tests/test_video_ingest.py
@@ -388,3 +388,77 @@ async def test_ingest_from_url_empty_chunks_returns_stored_no_chunks():
     data = response.json()
     assert data["status"] == "stored_no_chunks"
     assert data["chunks_created"] == 0
+
+
+async def test_ingest_from_url_embedding_failure_cleans_up_orphan_video():
+    """embed_batch raising → video row is deleted and 502 returned (no orphan row)."""
+    mock_video = {
+        "id": "v-from-url-1",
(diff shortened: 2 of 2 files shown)


coleam00:

Textbook execution of a small, well-scoped fix. The production change is a 2-line cleanup call inserted in exactly the right place, with a comment copied verbatim from the sibling `/ingest` handler so the two embedding-failure branches now read identically. The regression test patches the correct import boundaries, drives `embed_batch` to raise, and asserts both the 502 response and that `delete_video` was awaited once with the created video id, plus a defensive assertion that `create_chunk` was not called. The implementation faithfully delivers everything the plan specified with no scope creep — no touching the out-of-scope 500 branch, no refactor to deduplicate the cleanup logic, no unrelated edits. The only quibble is that neither handler wraps `delete_video` in a try/except, so a failure inside cleanup would mask the original 502 — but that's a pre-existing pattern in `/ingest` and consistency is the right call here.

## Scores

|---|---|---|
| solves_issue | 10/10 | Adds `await repository.delete_video(video_id)` in the 502 branch of `/ingest/from-url` exactly as the issue specified, and adds the requested regression test. Mirrors `/ingest` precisely. |
| root_cause | 10/10 | Fixes the true root cause: the missing cleanup call after a row is created but before chunks/embeddings exist. Not a workaround — it mirrors the established pattern from the sibling handler. |
| scope_discipline | 10/10 | Surgical: 2 lines of production code (comment + await delete) plus one new test. No drive-by refactors, no touching the out-of-scope 500 branch, no unrelated edits. |
| subtle_correctness | 9/10 | Async/await correct; uses the same comment text as `/ingest`; preserves the `raise HTTPException(...) from exc` chain. delete_video runs before the raise, matching the sibling handler. Minor: doesn't wrap delete in try/except (if delete itself fails the original 502 is lost), but `/ingest` doesn't either, so consistency wins. |
| code_quality | 10/10 | Reads identically to the `/ingest` handler — same comment, same call shape, same ordering. Maximally idiomatic to this codebase since it literally copies the established pattern. |
| test_discipline | 10/10 | New focused regression test exercises the embedding-failure path, asserts 502, asserts `delete_video` awaited once with the correct id, and defensively asserts `create_chunk` was not awaited. Patches at the right import boundaries; no existing tests gamed. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan line-for-line: same comment text, same call placement, same test name, same assertions, same patch targets. Zero drift. |

---

PR #235, opened 2026-05-21

### LINKED ISSUE #229: fix(rag): hybrid retrieval is non-deterministic on tied scores
## Summary

## Current behavior

## Acceptance

## Out of scope

coleam00:
Fixes #229

**Source run-id:** `ec25c17e-7669-4615-95b1-2bba8d662624`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -280,7 +280,7 @@ async def keyword_search(
             FROM chunks
             WHERE search_vector @@ plainto_tsquery($1)
               AND source_type = ANY($3::text[])
-            ORDER BY rank DESC
+            ORDER BY rank DESC, id ASC
             LIMIT $2
             """,
             query,
@@ -326,7 +326,7 @@ async def vector_search_pg(
                    embedding::vector <=> $1::vector AS distance
             FROM chunks
             WHERE source_type = ANY($3::text[])
-            ORDER BY distance
+            ORDER BY distance ASC, id ASC
             LIMIT $2
             """,
             embedding_json,
--- a/app/backend/rag/retriever_hybrid.py
+++ b/app/backend/rag/retriever_hybrid.py
@@ -196,5 +196,5 @@ def _rrf_merge(
         if chunk_id not in rows:
             rows[chunk_id] = row
 
-    ranked = sorted(scores, key=scores.__getitem__, reverse=True)[:top_k]
+    ranked = sorted(scores, key=lambda cid: (-scores[cid], cid))[:top_k]
     return [{**rows[cid], "rrf_score": scores[cid]} for cid in ranked]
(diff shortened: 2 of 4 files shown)


coleam00:

An exemplary small fix: precisely scoped, structurally correct, and matched by tests. Determinism is enforced at both the SQL and Python layers, which is the right defense-in-depth approach since either alone would still leave a non-deterministic seam (Postgres tie order or RRF tied-score order). The lambda-based RRF sort key is a clean improvement over the previous `__getitem__` form, the lexicographic tiebreak by TEXT chunk id is well-defined for the schema, and the new regression test would genuinely catch a future regression.

Minor weaknesses: the keyword_search SQL test wasn't tightened to assert the new `id ASC` secondary sort the way the vector_search_pg test was, so half of the SQL change is unverified at the assertion level. Beyond that, plan-to-implementation fidelity is essentially perfect.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly addresses the non-determinism: adds deterministic secondary sort (`id ASC`) to both keyword_search and vector_search_pg, and breaks ties in _rrf_merge by chunk id. Identical inputs now yield identical orderings. |
| root_cause | 10/10 | Fixes both layers where non-determinism could leak in — SQL tie ordering (Postgres has no inherent stable order) and the final RRF sort. Defense-in-depth rather than a single bandaid. |
| scope_discipline | 10/10 | Surgical: 4 files, 30 additions / 5 deletions. Only the two SQL queries, the RRF sort, and the matching tests are touched. No incidental refactoring, no out-of-scope changes. |
| subtle_correctness | 9/10 | Tuple key `(-scores[cid], cid)` correctly inverts score for descending order while keeping cid ascending — avoids the `reverse=True` foot-gun of flipping both. Chunk ids are TEXT so lexicographic order is well-defined. Distance is non-negative so `ORDER BY distance ASC` matches the prior implicit ASC behavior. |
| code_quality | 9/10 | One-line lambda is idiomatic Python and clearer than the previous `__getitem__` trick. SQL changes are minimal and consistent. Comment update on the existing test accurately reflects the new behavior. |
| test_discipline | 9/10 | Adds a focused regression test (`test_rrf_merge_deterministic_with_tied_scores`) that constructs an actual tied-score scenario and asserts lexicographic ordering — would fail before the fix. Updates the SQL assertion test to match the new ORDER BY. The existing keyword_search assertion test wasn't tightened to match the new `id ASC` clause, a minor gap. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan item-for-item: both SQL changes, the lambda key, the new regression test, the comment fix on the existing 'tie-break by insertion order' test, and the updated `test_vector_search_pg_orders_by_distance_ascending` assertion. No drift in either direction. |

---

PR #236, opened 2026-05-21
[bench-OO] fix(rag): make hybrid retrieval deterministic on tied scores

### LINKED ISSUE #229: fix(rag): hybrid retrieval is non-deterministic on tied scores

coleam00:
Fixes #229

**Source run-id:** `ef0a8938-31c6-407c-9766-05660d827243`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -280,7 +280,7 @@ async def keyword_search(
             FROM chunks
             WHERE search_vector @@ plainto_tsquery($1)
               AND source_type = ANY($3::text[])
-            ORDER BY rank DESC
+            ORDER BY rank DESC, id ASC
             LIMIT $2
             """,
             query,
@@ -326,7 +326,7 @@ async def vector_search_pg(
                    embedding::vector <=> $1::vector AS distance
             FROM chunks
             WHERE source_type = ANY($3::text[])
-            ORDER BY distance
+            ORDER BY distance ASC, id ASC
             LIMIT $2
             """,
             embedding_json,
--- a/app/backend/rag/retriever_hybrid.py
+++ b/app/backend/rag/retriever_hybrid.py
@@ -196,5 +196,5 @@ def _rrf_merge(
         if chunk_id not in rows:
             rows[chunk_id] = row
 
-    ranked = sorted(scores, key=scores.__getitem__, reverse=True)[:top_k]
+    ranked = sorted(scores, key=lambda cid: (-scores[cid], cid))[:top_k]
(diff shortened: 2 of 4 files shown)


coleam00:

A near-textbook execution of a small determinism fix. The PR closes the bug at every layer the issue identified (both SQL ORDER BY clauses and the in-memory RRF sort), leans on the existing unique TEXT primary key so the tiebreak is provably stable without any schema work, and ships a regression test that explicitly verifies stability across input permutations rather than just smoke-testing one ordering.

Strengths: tight scope, true root-cause fix, surgical diff, and high plan-impl fidelity — the implementation reads like a direct transcription of the plan's `Specific changes` list. The only nit is the slightly less readable `lambda cid: (-scores[cid], cid)` compared to the prior `scores.__getitem__`, but that's the right tradeoff for correctness in a single sort call. Nothing meaningfully weak here.

## Scores

|---|---|---|
| solves_issue | 10/10 | Addresses every layer the issue called out: keyword_search ORDER BY, vector_search_pg ORDER BY, and the _rrf_merge final sort all gain stable tiebreakers. Identical inputs now produce identical ordering. |
| root_cause | 10/10 | Fixes the actual source of non-determinism (Postgres ORDER BY with no unique tiebreaker, and dict-key sort with no secondary key), not a symptom. Uses the existing unique TEXT primary key `id`/`chunk_id`, so the tiebreak is guaranteed unique without schema work. |
| scope_discipline | 10/10 | Exactly 4 files (2 source, 2 test), 39 additions / 6 deletions, no drive-by edits. Stays out of retriever.py, expansion.py, and orchestration as the plan required. |
| subtle_correctness | 10/10 | The `sorted(scores, key=lambda cid: (-scores[cid], cid))` pattern correctly preserves DESC-by-score while making the tiebreak ASC-by-id in a single pass. SQL tiebreakers use the chunks primary key which is unique. No async/typing footguns introduced. |
| code_quality | 9/10 | Minimal, idiomatic, matches existing style. The lambda is a small readability hit vs. the prior `scores.__getitem__`, but it's the cleanest way to encode the composite key; updated docstrings and tie-break comment keep documentation honest. |
| test_discipline | 10/10 | Adds a real regression test that exercises permutation stability AND asserts the specific lexicographic order, plus updates the two SQL substring assertions to require the new tiebreaker. Tests would fail if a future refactor dropped either layer. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan one-for-one: same four files, same line-level changes, the exact `(-scores[cid], cid)` lambda the plan specified, and the named regression test `test_rrf_merge_breaks_ties_deterministically`. No drift in either direction. |

---

PR #237, opened 2026-05-21
[bench-KK] review: remove accidental lock file, add to gitignore

### LINKED ISSUE #229: fix(rag): hybrid retrieval is non-deterministic on tied scores

coleam00:
Fixes #229

**Source run-id:** `cd6d5c38-4cb1-4a59-89b8-f7945ba937f8`


coleam00:

A clean, high-quality fix that does exactly what the issue asks for and faithfully executes the plan. Both layers of the non-determinism (SQL row order on ties and RRF score-tie ordering) get explicit tiebreakers, and the regression test pins both that runs are stable and that the tie resolves the expected way (chunk-id ASC). The only blemish is a small scope-discipline lapse: the self-review commit bundles a `.gitignore` addition for `.pi-loop.json.lock` that is unrelated to issue #229 and per CLAUDE.md should have been a separate PR. Otherwise the change is surgical, idiomatic, well-tested, and structurally correct.

## Scores

|---|---|---|
| solves_issue | 10/10 | Adds the exact tiebreakers the issue asked for: `id ASC` secondary sort in both `keyword_search` and `vector_search_pg`, plus a composite-key deterministic tiebreak in `_rrf_merge`'s final sort. Acceptance criteria are met. |
| root_cause | 10/10 | Fixes non-determinism at both layers it originates from — the SQL ordering (Postgres's row-order ambiguity on ties) and the Python RRF sort. Defense-in-depth, not a band-aid. |
| scope_discipline | 8/10 | Surgical changes to the three files named in the plan plus a targeted test. Lightly bundles an unrelated `.gitignore` entry for `.pi-loop.json.lock` in a 'self-review' cleanup commit — minor scope creep against CLAUDE.md's one-issue-per-PR rule. |
| subtle_correctness | 9/10 | Composite key `(-scores[cid], cid)` correctly maps reverse-sort-by-score into ascending order while keeping ID tiebreak ascending. SQL tiebreakers are stable (id is the table PK). Test math (1/60+1/61 == 1/61+1/60) is sound. |
| code_quality | 9/10 | Idiomatic, minimal, matches surrounding style. Test docstring + inline comments explain the construction. Only nit: the new test's inline chunk dicts could share a helper like `_make_chunk`, but that's cosmetic. |
| test_discipline | 9/10 | Targeted regression test that constructs a genuine tied-score case and runs `_rrf_merge` 10× to assert stable ordering, plus pins the expected order (`cx` before `cy`) so a future regression to position-based tiebreak would fail loudly. No existing tests gamed. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan line-for-line: same files, same SQL edits, same composite-key replacement, same regression test name and construction strategy. Zero drift. |

---

PR #238, opened 2026-05-21

### LINKED ISSUE #228: fix(ingest): /ingest/from-url leaves an orphan video row when embedding fails

coleam00:
Fixes #228

**Source run-id:** `af8db31f-0aa1-4873-ae1a-341a7d3fe6de`


coleam00:

A clean, surgical bug fix that exactly executes both the issue and the plan. The route change is a single-line cleanup mirroring the sibling /ingest handler — same comment, same ordering, same semantics — which is the correct structural fix rather than a workaround. The regression test is well-scoped: it mocks the boundary (fetch_video_for_ingest, create_video, chunk_video_fallback, embed_batch, delete_video), exercises the real route via ASGITransport, and asserts the 502 status, error detail, and that delete_video was awaited with the created video_id. Weaknesses are minor: the test could additionally assert call ordering (delete before raise) and the handler does not guard against delete_video itself raising, though this matches the existing /ingest pattern and is appropriately out of scope. Plan-to-impl fidelity is essentially perfect.

## Scores

|---|---|---|
| solves_issue | 10/10 | Adds the exact missing `await repository.delete_video(video_id)` call in the /ingest/from-url 502 branch as the issue requested, and adds a regression test asserting the orphan-cleanup behavior. |
| root_cause | 10/10 | Directly fixes the missing cleanup at the failure site, mirroring the proven /ingest pattern. No band-aid — it removes the orphan-row condition rather than masking it. |
| scope_discipline | 10/10 | Two-file diff: a single 2-line addition in the route plus one new isolated test. No unrelated edits, no touching the 500 branch, no refactors. |
| subtle_correctness | 9/10 | Correctly awaits delete_video before raising 502; matches existing handler semantics. Minor nit: doesn't guard against delete_video itself raising (neither does the sibling handler), so behavior is consistent. |
| code_quality | 9/10 | Idiomatic, mirrors sibling pattern verbatim including the inline comment. Test uses the codebase's existing AsyncMock/patch style and ASGITransport pattern consistently. |
| test_discipline | 9/10 | New targeted regression test asserts 502, the embed call fired, create_video was called, and delete_video was awaited with the correct video_id. Solid mocked-boundary test; could additionally assert call order, but coverage of the bug is meaningful. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan precisely: single-line cleanup added in the named except block, regression test added with the exact name and assertions described, no out-of-scope changes. |

---

PR #239, opened 2026-05-21
[bench-OK] review: address self-review findings

### LINKED ISSUE #228: fix(ingest): /ingest/from-url leaves an orphan video row when embedding fails

coleam00:
Fixes #228

**Source run-id:** `16ed26ab-5bf6-4d6b-82ef-a90136ffe260`


coleam00:

Textbook execution of a small, well-scoped bug fix. The diff is exactly the 2-line cleanup-call addition the issue called for, in the right place, with the right wording matching the sibling `/ingest` handler. The regression test is thorough — it patches the full call chain, asserts the 502 response, the precise `delete_video(video_id)` await, and verifies that downstream side effects (chunk creation, cache invalidation) correctly never fire on the failure path. Plan fidelity is essentially perfect: every bullet from `source-plan.md` (including the negative assertions on `invalidate_cache` / `invalidate_catalog`) is reflected in the implementation. The only nit is that, like the sibling handler, it doesn't defend against `delete_video` itself raising — but faithfully mirroring existing behavior is the right call here per scope discipline.

## Scores

|---|---|---|
| solves_issue | 10/10 | Adds `await repository.delete_video(video_id)` in the 502 branch of `/ingest/from-url`, exactly mirroring `/ingest` as the issue requested, and adds a regression test asserting no orphan remains. |
| root_cause | 10/10 | Fix is applied at the exact site of the leak — between the failure log and the raise — addressing the structural inconsistency between the two handlers rather than papering over it elsewhere. |
| scope_discipline | 10/10 | Touches only `routes/ingest.py` (2 lines) and `tests/test_video_ingest.py` (one new test). No refactors, no helper extraction, no changes to the 500 branch — strict adherence to issue scope. |
| subtle_correctness | 9/10 | Correctly awaits the async cleanup, preserves `from exc` chain, mirrors the comment wording. Tests also verify negatives (create_chunk not awaited, cache invalidators not called). Minor: doesn't guard against `delete_video` itself raising, but neither does the sibling handler, so it's faithful. |
| code_quality | 10/10 | Two-line change is idiomatic, comment matches the sibling handler verbatim, test is well-structured with clear naming and explicit positive+negative assertions. |
| test_discipline | 10/10 | Adds a focused regression test that drives the failure path end-to-end via ASGITransport, asserts 502, exact `delete_video` call, and that downstream side effects didn't happen. No existing tests modified to pass. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan line-for-line: same comment wording, same insertion point, test name `test_ingest_from_url_embedding_failure_cleans_up_video` and all enumerated patches/assertions (including the negative assertions for cache invalidators) are present. |

---

PR #240, opened 2026-05-21
[bench-OK] review: remove accidentally committed process lock file

### LINKED ISSUE #229: fix(rag): hybrid retrieval is non-deterministic on tied scores

coleam00:
Fixes #229

**Source run-id:** `ddad2e6b-b604-498a-8197-e6fd1d10f8b0`


coleam00:

A textbook surgical bug fix. The diff targets exactly the two determinism sources called out in issue #229 — SQL ORDER BY ties and Python dict-insertion-order leakage through `_rrf_merge` — using the chunk `id` PK as a stable, content-derived tiebreaker at both layers. Scope discipline is excellent (4 files, +19/-4), the regression test meaningfully exercises the tied-score path by swapping input order, and the implementation matches the plan line-for-line. The only minor nit is that the test suite could include one more case covering full RRF ordering under mixed ties, but the existing test already proves the core invariant. Strong execution from the mixed Opus-plan/Kimi-impl cell.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly addresses the issue's two root causes: adds `id ASC` tiebreakers to both `keyword_search` and `vector_search_pg`, and replaces the RRF sort with a composite key `(-score, cid)` so dict insertion order can no longer leak. Acceptance criteria (deterministic ordering, stable secondary tiebreaker, regression test) all satisfied. |
| root_cause | 10/10 | Fixes determinism at both layers identified in the issue — SQL-level ties and the post-merge Python sort — rather than masking either. The chunk `id` PK is a content-derived stable key with no extra schema cost. |
| scope_discipline | 10/10 | Four files, +19/-4. Touches only the two ORDER BY clauses, the one `sorted()` call, and tightens/adds tests for exactly those changes. Nothing extraneous. |
| subtle_correctness | 9/10 | Tiebreak uses ascending text PK, which is consistent across SQL and Python paths. `lambda cid: (-scores[cid], cid)` correctly inverts score for descending while keeping cid ascending. Minor caveat: relies on `id` being non-null and unique, which it is as the PK. |
| code_quality | 9/10 | Idiomatic Python and SQL; the new lambda is slightly less terse than `scores.__getitem__` but is the standard pattern for compound keys. No new comments needed because intent is clear from the change. |
| test_discipline | 9/10 | Adds a genuine regression test (`test_rrf_merge_breaks_ties_deterministically`) that swaps input order and asserts identical output, which would fail under the old behavior. Also tightens SQL substring asserts to lock in the tiebreaker. Could optionally have added a third test for full RRF ordering with mixed ties, but coverage is adequate. |
| plan_impl_fidelity | 10/10 | Implementation is a verbatim execution of the plan: same two ORDER BY edits, same `sorted` rewrite, same two test additions, no drift in or out of scope. |

---

PR #248, opened 2026-05-31
[bench-KK] review: remove committed lock file and add to .gitignore

### LINKED ISSUE #242: fix(rag): keyword_search ignores its language param, query regconfig drifts from index

```python
    ...
    ...
```

### Acceptance

### Out of scope

### Why this tests reasoning

coleam00:
Fixes #242

**Source run-id:** `72013521-2c6c-48ca-a862-ca6b92ef2564`


coleam00:
Superseded: re-running wave 1 against a base that gitignores the Pi .pi-loop.json.lock artifact (was biasing scope).

PR #252, opened 2026-05-31

### LINKED ISSUE #242: fix(rag): keyword_search ignores its language param, query regconfig drifts from index

coleam00:
Fixes #242

**Source run-id:** `38770889-3c70-42e9-8439-ff8795da32a9`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -276,9 +276,9 @@ async def keyword_search(
         rows = await conn.fetch(
             """
             SELECT id, video_id, content, chunk_index, start_seconds, end_seconds, snippet,
-                   ts_rank(search_vector, plainto_tsquery($1)) AS rank
+                   ts_rank(search_vector, plainto_tsquery($4::regconfig, $1)) AS rank
             FROM chunks
-            WHERE search_vector @@ plainto_tsquery($1)
+            WHERE search_vector @@ plainto_tsquery($4::regconfig, $1)
               AND source_type = ANY($3::text[])
             ORDER BY rank DESC
             LIMIT $2
@@ -286,6 +286,7 @@ async def keyword_search(
             query,
             top_k,
             allowed_source_types,
+            language,
         )
     return [dict(r) for r in rows]
 
--- a/app/backend/tests/test_repository_hybrid_search.py
+++ b/app/backend/tests/test_repository_hybrid_search.py
@@ -49,16 +49,44 @@ class TestKeywordSearch:
         assert "search_vector" in sql
         assert "@@" in sql  # tsvector match operator
 
+        # Both plainto_tsquery calls must carry the regconfig parameter so
+        # query-side tokenization matches the index-side to_tsvector('english').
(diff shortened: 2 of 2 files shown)


coleam00:

This is a textbook-correct fix. The OO (plan=Opus, impl=Opus) cell correctly diagnosed that the real defect is a silently dropped parameter and bound `language` as `$4::regconfig` into both plainto_tsquery calls — the ts_rank expression and the WHERE @@ clause — so query-time tokenization can never drift from the generated column's hardcoded 'english' index. It deliberately avoided both wrong fixes the issue flagged (mutating default_text_search_config, or patching only one clause), and stayed entirely within scope: no migration, pool-init, or pgvector changes. Parameter ordering is correct and the ::regconfig cast is valid.

Test discipline is strong: the new test exercises a non-default language and both the updated and added tests assert exactly two `plainto_tsquery($4::regconfig, $1)` occurrences, explicitly guarding the partial-fix regression. Fidelity to the plan is essentially perfect. The only reason a couple of dimensions sit at 9 rather than 10 is that the change is small enough that there is little to distinguish exceptional from merely solid execution — there are no novel edge cases handled beyond the requirement, which is appropriate for the issue's scope.

## Scores

|---|---|---|
| solves_issue | 10/10 | Binds `language` as `$4::regconfig` into BOTH plainto_tsquery calls (ts_rank and WHERE) and appends it to conn.fetch, exactly satisfying the acceptance criteria so query-side regconfig always matches the index's hardcoded 'english'. |
| root_cause | 10/10 | Fixes the true root cause — the dropped parameter — rather than the tempting band-aids the issue warned against (setting default_text_search_config at the pool, or regconfig on only one clause). Query and index regconfig now stay identical regardless of the DB GUC. |
| scope_discipline | 10/10 | Surgical: a 4-line production edit in repository.py plus targeted test updates. No migration, pool-init, GUC, or vector_search_pg changes — respects every out-of-scope boundary. |
| subtle_correctness | 9/10 | Parameter ordering is correct ($1=query, $2=top_k, $3=allowed_source_types, $4=language) with language appended last, and the ::regconfig cast on the text param is valid Postgres. Both calls updated, avoiding the ranking-drift foot-gun. No async/type issues introduced. |
| code_quality | 9/10 | Idiomatic to the asyncpg/$N positional style of the repo, minimal and readable, with a clear inline comment in tests explaining why both calls need the regconfig. Signature and return shape untouched. |
| test_discipline | 9/10 | Adds a meaningful guard test (language='simple' reaching both calls) plus strengthens the existing test to assert exactly two `plainto_tsquery($4::regconfig, $1)` occurrences — directly guarding against the partial-fix failure mode. Tests assert behavior, not gamed to pass. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan verbatim: $4::regconfig on both lines, language appended as the final fetch arg, no signature change, and both planned tests (updated + new) delivered with the asserted indices and counts. |

---

PR #253, opened 2026-05-31

### LINKED ISSUE #242: fix(rag): keyword_search ignores its language param, query regconfig drifts from index

coleam00:
Fixes #242

**Source run-id:** `35ad30bf-34c7-4fef-9dae-4a9760733d3a`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -276,9 +276,9 @@ async def keyword_search(
         rows = await conn.fetch(
             """
             SELECT id, video_id, content, chunk_index, start_seconds, end_seconds, snippet,
-                   ts_rank(search_vector, plainto_tsquery($1)) AS rank
+                   ts_rank(search_vector, plainto_tsquery($4::regconfig, $1)) AS rank
             FROM chunks
-            WHERE search_vector @@ plainto_tsquery($1)
+            WHERE search_vector @@ plainto_tsquery($4::regconfig, $1)
               AND source_type = ANY($3::text[])
             ORDER BY rank DESC
             LIMIT $2
@@ -286,6 +286,7 @@ async def keyword_search(
             query,
             top_k,
             allowed_source_types,
+            language,
         )
     return [dict(r) for r in rows]
 
--- a/app/backend/tests/test_repository_hybrid_search.py
+++ b/app/backend/tests/test_repository_hybrid_search.py
@@ -51,14 +51,42 @@ class TestKeywordSearch:
 
         # Verify positional args: query string and top_k
         args = call_args[0]
-        # args is (sql_string, query_string, top_k_int)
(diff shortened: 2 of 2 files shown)


coleam00:

An exemplary small reasoning PR. It identifies and fixes the true root cause — the silently-dropped `language` parameter — by binding it into both plainto_tsquery calls as `$4::regconfig`, sidestepping the two tempting wrong fixes (setting the GUC at the pool, or updating only the WHERE clause and letting ts_rank drift). The regconfig cast shows correct understanding of asyncpg's text binding and Postgres FTS overload resolution. Scope is surgical (two files, +32/-3) and stays inside every out-of-scope guardrail. Tests are meaningful and non-gamed: a new spanish-language case plus tightened assertions that pin both occurrences and the absence of the bare form. The only reason it isn't a perfect 70 is that, by the nature of a mocked-boundary unit test, correctness against an actual non-english `default_text_search_config` is asserted at the SQL-string level rather than verified end-to-end against Postgres — a reasonable and CLAUDE.md-compliant limitation, but it leaves a sliver of behavioral verification to the manual tier. Implementation faithfully executes a clear, correct plan.

## Scores

|---|---|---|
| solves_issue | 10/10 | Binds `language` into both plainto_tsquery calls (ts_rank and WHERE) as `$4::regconfig`, exactly satisfying the acceptance criteria so query-side regconfig matches the hardcoded english index regardless of default_text_search_config. |
| root_cause | 10/10 | Fixes the true root cause — the dropped parameter — rather than papering over it by setting the GUC at the pool level (the explicitly-rejected wrong fix). Query and index regconfig now stay identical. |
| scope_discipline | 10/10 | Touches only repository.py (the SQL + fetch args) and its test file; +32/-3, two files. No drift into the migration, postgres.py, or vector_search_pg as the out-of-scope list demanded. |
| subtle_correctness | 9/10 | Nails the two subtle foot-guns: updates BOTH plainto_tsquery occurrences (so ranking doesn't drift), and casts `$4::regconfig` because asyncpg binds strings as text and the bare overload would error. Parameter positions left correct. |
| code_quality | 9/10 | Idiomatic to the codebase — parameterized `$N` placeholders, SQL confined to repository.py, follows the existing positional-arg pattern. Readable and minimal with no stylistic noise. |
| test_discipline | 9/10 | Adds a dedicated test binding language="spanish" plus strengthens the existing test to assert exactly two `plainto_tsquery($4::regconfig, $1)` and that no bare `plainto_tsquery($1)` remains. Targeted and not gamed; mocked at the boundary per CLAUDE.md. |
| plan_impl_fidelity | 10/10 | The diff matches the plan item-for-item: both clauses changed to `$4::regconfig`, `language` appended to conn.fetch, and the test asserts the 4th-arg binding plus both SQL occurrences. No drift in either direction. |

---

PR #254, opened 2026-05-31

### LINKED ISSUE #242: fix(rag): keyword_search ignores its language param, query regconfig drifts from index

coleam00:
Fixes #242

**Source run-id:** `58599f0f-2663-4449-aa9a-dd008c19dfe5`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -276,9 +276,9 @@ async def keyword_search(
         rows = await conn.fetch(
             """
             SELECT id, video_id, content, chunk_index, start_seconds, end_seconds, snippet,
-                   ts_rank(search_vector, plainto_tsquery($1)) AS rank
+                   ts_rank(search_vector, plainto_tsquery($4::regconfig, $1)) AS rank
             FROM chunks
-            WHERE search_vector @@ plainto_tsquery($1)
+            WHERE search_vector @@ plainto_tsquery($4::regconfig, $1)
               AND source_type = ANY($3::text[])
             ORDER BY rank DESC
             LIMIT $2
@@ -286,6 +286,7 @@ async def keyword_search(
             query,
             top_k,
             allowed_source_types,
+            language,
         )
     return [dict(r) for r in rows]
 
--- a/app/backend/tests/test_repository_hybrid_search.py
+++ b/app/backend/tests/test_repository_hybrid_search.py
@@ -14,7 +14,7 @@ class TestKeywordSearch:
     """Tests for keyword_search() SQL execution."""
 
     async def test_keyword_search_calls_fetch_with_correct_params(self):
-        """keyword_search passes query and top_k as $1, $2 to the SQL query."""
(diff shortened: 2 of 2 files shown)


coleam00:

An exemplary fix for a reasoning-trap issue. The implementation correctly identifies that the dropped `language` parameter must be bound to BOTH plainto_tsquery call sites — sidestepping both tempting wrong answers (setting the GUC at the pool level, or fixing only the WHERE clause and letting ts_rank drift). The `$4::regconfig` cast is the correct, idiomatic Postgres approach, positional argument ordering is right, and scope is perfectly disciplined: only the SQL/args and the test file changed, with every out-of-scope file left untouched. Test coverage is meaningful and targeted, asserting the regconfig binding appears exactly twice and adding a dedicated non-default-language case. The only minor weakness is that tests mock at the connection boundary, so they verify the SQL string and argument wiring rather than actual cross-config behavior in a live Postgres — acceptable given repo conventions, but it means the 'results identical across default_text_search_config' guarantee is proven by construction rather than by an integration test. Plan-to-implementation fidelity is essentially perfect.

## Scores

|---|---|---|
| solves_issue | 10/10 | Binds `language` as `$4::regconfig` into BOTH the `ts_rank` and `WHERE @@` plainto_tsquery calls and appends `language` to the fetch args — exactly the acceptance criteria. Query-side regconfig now matches the hardcoded index-side 'english' tsvector regardless of the connection's default_text_search_config. |
| root_cause | 10/10 | True root cause: the parameter was declared but never bound. The fix wires it through to both query sites rather than papering over it by setting default_text_search_config at the pool level (the explicitly-warned-against band-aid). |
| scope_discipline | 10/10 | Surgical — only `repository.py` (the SQL + arg list) and its test file. No migration, postgres.py, vector_search_pg, or retriever_hybrid.py touched, matching the out-of-scope list precisely. |
| subtle_correctness | 9/10 | Correctly fixes BOTH plainto_tsquery sites (the tempting half-fix touches only `@@` and lets ts_rank drift). The `$4::regconfig` cast is necessary since asyncpg sends text, and positional arg order ($3 source_types, $4 language) aligns with the SQL. No async/type foot-guns introduced. |
| code_quality | 9/10 | Clean, idiomatic, parameterized per CLAUDE.md SQL conventions; the regconfig cast is the standard Postgres idiom. Minimal and readable with no churn. |
| test_discipline | 9/10 | Updated the existing test to assert `plainto_tsquery($4::regconfig, $1)` appears exactly twice and verifies args[3]/args[4], plus added a focused `test_keyword_search_uses_custom_language` for the non-default path. Tests are mocked at the conn boundary (no real cross-config DB behavior verified), but that matches repo conventions. |
| plan_impl_fidelity | 10/10 | The diff executes the plan item-for-item: both plainto_tsquery rewrites, the appended `language` arg, the updated default-language assertions, and the new custom-language test — no drift, no unplanned additions. |

---

PR #255, opened 2026-05-31

### LINKED ISSUE #242: fix(rag): keyword_search ignores its language param, query regconfig drifts from index

coleam00:
Fixes #242

**Source run-id:** `62e4d518-e29e-4f0d-97c4-1d5590daae3c`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -276,9 +276,9 @@ async def keyword_search(
         rows = await conn.fetch(
             """
             SELECT id, video_id, content, chunk_index, start_seconds, end_seconds, snippet,
-                   ts_rank(search_vector, plainto_tsquery($1)) AS rank
+                   ts_rank(search_vector, plainto_tsquery($4::text::regconfig, $1)) AS rank
             FROM chunks
-            WHERE search_vector @@ plainto_tsquery($1)
+            WHERE search_vector @@ plainto_tsquery($4::text::regconfig, $1)
               AND source_type = ANY($3::text[])
             ORDER BY rank DESC
             LIMIT $2
@@ -286,6 +286,7 @@ async def keyword_search(
             query,
             top_k,
             allowed_source_types,
+            language,
         )
     return [dict(r) for r in rows]
 
--- a/app/backend/tests/test_repository_hybrid_search.py
+++ b/app/backend/tests/test_repository_hybrid_search.py
@@ -49,16 +49,53 @@ class TestKeywordSearch:
         assert "search_vector" in sql
         assert "@@" in sql  # tsvector match operator
 
-        # Verify positional args: query string and top_k
+        # Verify positional args: query string, top_k, allowed_source_types, language
         args = call_args[0]
-        # args is (sql_string, query_string, top_k_int)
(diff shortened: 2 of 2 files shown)


coleam00:

An excellent, high-confidence fix. It correctly identifies that the symptom (degraded keyword hits on non-english default_text_search_config) stems from a dropped parameter, and fixes it at the source by binding `language` into both plainto_tsquery calls with the same $4 placeholder so the @@ filter and ts_rank score can never diverge. The standout detail is the $4::text::regconfig double-cast, which navigates asyncpg's lack of a regconfig codec while still erroring loudly on an invalid config name — avoiding both the bind-time failure of a bare regconfig param and the silent NULL-fallback footgun of to_regconfig. Scope is perfectly disciplined and tests are meaningful and targeted, including a regression guard against the partial-fix (only-WHERE) trap the issue warned about.

The only minor gaps: the non-obvious double-cast lacks an inline comment for future maintainers, and the optional live-Postgres test proving GUC-independence (the actual behavioral guarantee) was not added — the coverage is mock-based, asserting SQL shape and bound args rather than real tokenization behavior, though the plan explicitly gated that as optional. Plan-to-impl fidelity is near-perfect. This is a strong cell result: a planner that reasoned correctly about Postgres FTS regconfig semantics and an implementer that executed it faithfully.

## Scores

|---|---|---|
| solves_issue | 10/10 | Binds `language` as the regconfig argument to BOTH plainto_tsquery calls (ts_rank and WHERE) using the same $4 placeholder, exactly the acceptance criteria. Query-side regconfig now always matches the english-pinned generated index regardless of default_text_search_config. |
| root_cause | 10/10 | Threads the dropped parameter through to the SQL — the true root cause — rather than papering over it by pinning the GUC at the pool level (the explicitly-rejected wrong fix). Index/query regconfig can no longer drift. |
| scope_discipline | 10/10 | Surgical: only repository.py (the one function) and its test file. No migration, pool-init, vector_search_pg, signature, or caller changes — respects every out-of-scope boundary. |
| subtle_correctness | 10/10 | The $4::text::regconfig double-cast is the sophisticated correct choice: asyncpg has no regconfig codec so a bare $4::regconfig would fail at bind time, and ::regconfig errors loudly on a bad config name (unlike to_regconfig which returns NULL and silently reintroduces the GUC fallback bug). Same $4 in both clauses prevents rank/filter drift. |
| code_quality | 9/10 | Clean, parameterized (no f-string SQL), idiomatic to the asyncpg style in this file. A one-line comment explaining the non-obvious ::text::regconfig double-cast would have helped future readers, but it's otherwise production-grade. |
| test_discipline | 9/10 | Added two targeted regression tests — regconfig present in both clauses, and language='simple' threaded through — plus updated existing assertions to verify the new bound arg. Directly guards the partial-fix regression. Only the optional live-Postgres GUC-independence test was omitted (plan marked it optional). |
| plan_impl_fidelity | 10/10 | Implementation executes the plan essentially verbatim: $4::text::regconfig added to both plainto_tsquery calls, language appended to the fetch tuple after allowed_source_types, signature/default untouched, and the planned both-clauses + language-threaded tests added. No drift. |

---

PR #256, opened 2026-05-31

### LINKED ISSUE #243: fix(integrations): Circle paid-membership check fails open when active field is absent

```python
if member is None:
    return False
    return False
```

### Acceptance

### Out of scope

### Why this tests reasoning

coleam00:
Fixes #243

**Source run-id:** `a9aec468-125b-4cd7-a978-f28adb528ac1`

### DIFF
--- a/app/backend/integrations/circle.py
+++ b/app/backend/integrations/circle.py
@@ -64,7 +64,7 @@ async def verify_paid_member(email: str) -> bool:
             member = _extract_member(r.json())
             if member is None:
                 return False
-            if not member.get("active", True):
+            if not member.get("active", False):
                 # Inactive members lose access immediately.
                 return False
             member_id = member.get("id")
--- a/app/backend/tests/test_circle.py
+++ b/app/backend/tests/test_circle.py
@@ -110,6 +110,15 @@ async def test_inactive_member_returns_false():
 
 
 @respx.mock
+async def test_member_without_active_field_fails_closed():
+    respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(
+        return_value=httpx.Response(200, json={"id": 999})
+    )
+
+    assert await circle.verify_paid_member("absent-active@example.com") is False
+
+
+@respx.mock
 async def test_records_wrapped_response_handled():
     """Some Circle endpoints wrap a single record in `records: [...]`."""
     respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(


coleam00:

A clean, well-scoped fix that correctly diagnoses and repairs the fail-open default in the Circle paid-membership gate. The one-line change from `member.get("active", True)` to `member.get("active", False)` satisfies every acceptance criterion, honors the documented fail-closed contract, and respects all out-of-scope constraints (no changes to `_extract_member`, the timeout, or the two-step flow). The accompanying respx test is a genuine, targeted regression test for the exact missing-field case. Scope discipline, test discipline, and plan/impl fidelity are exemplary — the diff is a verbatim execution of the planning artifact.

The weaknesses are minor and stylistic rather than functional: the inline comment 'Inactive members lose access immediately' was not updated even though the branch now also denies members whose `active` field is absent, which slightly obscures the fail-closed intent at the heart of this issue. The PR also does not engage with the issue's hint that `active` may live under a different key in the search-result shape (a potential false-deny regression), though addressing that would have pushed into out-of-scope parsing logic. Net: a high-quality, correctly-reasoned, plan-faithful fix scoring 61/70.

## Scores

|---|---|---|
| solves_issue | 9/10 | Changing `member.get("active", True)` to `member.get("active", False)` exactly satisfies all three acceptance criteria: an absent active field now returns False, an explicitly active member still proceeds, and the access-group check is untouched. Minor reservation: it doesn't address the issue's aside about active status living under a different key in the search shape, though that is arguably out of scope. |
| root_cause | 8/10 | The literal root cause is the fail-open default value `True`; flipping it to `False` is a true structural fix aligned with the module's documented fail-closed contract, not a band-aid. Slightly short of 10 because the deeper concern (checking `active` on a search-result shape that may omit it) is left to the existing two-step flow, which was correctly out of scope. |
| scope_discipline | 10/10 | Surgical: one character-level change in circle.py plus one focused test. `_extract_member`, the timeout, and the two-step request flow are all left untouched, exactly honoring the out-of-scope list. |
| subtle_correctness | 8/10 | Logic is correct across all three cases (absent -> deny, True -> proceed, False -> deny); no async/await or type foot-guns introduced. Docked because the inline comment 'Inactive members lose access immediately' now also silently covers absent-field members and was not updated to reflect the new fail-closed-on-absence semantics. |
| code_quality | 7/10 | Minimal and idiomatic to the codebase. The change is the whole point about a fail-closed contract, yet the stale comment was not revised to explain that absence now denies; a brief 'fail closed: absent active field denies access' would have made intent explicit. Otherwise clean. |
| test_discipline | 9/10 | Adds a targeted regression test `test_member_without_active_field_fails_closed` mocking a search response with no `active` key (`{"id": 999}`) and asserting False. It reuses the existing respx pattern, directly exercises the bug, and is not a gamed/loosened test. |
| plan_impl_fidelity | 10/10 | source-plan.md was readable and the implementation executes it faithfully and completely: the planned one-line `True`->`False` flip on circle.py line 67 and the planned `test_member_without_active_field_fails_closed` test (200 response with `{"id": 999}`, asserting False) were both delivered verbatim, with all listed out-of-scope items respected. No drift. |

---

PR #257, opened 2026-05-31

### LINKED ISSUE #243: fix(integrations): Circle paid-membership check fails open when active field is absent

coleam00:
Fixes #243

**Source run-id:** `1a3a22e7-c337-453f-b8ad-f824e4e0128c`

### DIFF
--- a/app/backend/integrations/circle.py
+++ b/app/backend/integrations/circle.py
@@ -64,8 +64,9 @@ async def verify_paid_member(email: str) -> bool:
             member = _extract_member(r.json())
             if member is None:
                 return False
-            if not member.get("active", True):
-                # Inactive members lose access immediately.
+            if not member.get("active", False):
+                # Inactive members, or members whose status is absent/unknown,
+                # lose access (fail-closed).
                 return False
             member_id = member.get("id")
             if not member_id:
--- a/app/backend/tests/test_circle.py
+++ b/app/backend/tests/test_circle.py
@@ -110,6 +110,26 @@ async def test_inactive_member_returns_false():
 
 
 @respx.mock
+async def test_member_missing_active_field_returns_false():
+    """A member object without an explicit active key must be denied (fail-closed)."""
+    respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(
+        return_value=httpx.Response(200, json={"id": 999})
+    )
+
+    assert await circle.verify_paid_member("unknown@example.com") is False
+
+
+@respx.mock
+async def test_records_wrapped_missing_active_field_returns_false():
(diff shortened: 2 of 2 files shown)


coleam00:

A clean, high-quality fix that correctly identifies and repairs the true root cause: member.get("active", True) defaulted to granting access on a missing field, contradicting the module's fail-closed contract. Changing the default to False makes an absent active signal deny, satisfying every acceptance criterion while leaving explicitly-active members and the downstream access-group check intact. Scope discipline is exemplary — one logical line of production code, all out-of-scope items respected — and the comment is updated to document the invariant. Tests are targeted and meaningful, covering both the bare and records-wrapped response shapes for the missing-active case without disturbing existing assertions.

The only sliver of room for improvement is that the issue hinted Circle may carry status under a different key in the search-result shape; the fix does not probe alternate keys, so such a member would now be denied. This is the conservative, fail-closed behavior and is explicitly outside the stated acceptance scope, so it is not a defect — merely the boundary of what was asked. Implementation is a faithful execution of the plan.

## Scores

|---|---|---|
| solves_issue | 10/10 | Flips member.get("active", True) to default False so a member object lacking an active key is denied, exactly satisfying all three acceptance criteria; explicitly-active members still proceed and the access-group check is untouched. |
| root_cause | 10/10 | Fixes the actual defect — the wrong default value in the .get() call — rather than patching a symptom, and correctly connects the module's documented fail-closed contract to the absent-field case. The comment is updated to explain the invariant. |
| scope_discipline | 10/10 | Touches only circle.py (one logical line) and its test file; respects every out-of-scope boundary (_extract_member, timeout, two-step flow all untouched). 23/+2 lines, almost entirely tests. |
| subtle_correctness | 9/10 | Correctly reasons that for a fail-closed ACL, absence must deny — the conservative choice. Handles both bare and records-wrapped shapes. Minor: status under an alternate key (a possibility the issue raised) is not probed, but that is explicitly out of the acceptance scope and consistent with fail-closed. |
| code_quality | 9/10 | Idiomatic, minimal, and the replacement comment clearly documents the fail-closed rationale. Reads cleanly against the surrounding code. |
| test_discipline | 9/10 | Two meaningful, well-named regression tests cover the bare and records-wrapped missing-active shapes; existing active/inactive tests are preserved and not gamed. A community_members-wrapped variant would have been a nice extra but coverage is solid. |
| plan_impl_fidelity | 10/10 | The plan specified exactly this single-line default fix plus two tests (bare member + records-wrapped, both missing active), keeping existing tests passing. The diff delivers precisely that with no drift. |

---

PR #258, opened 2026-05-31

### LINKED ISSUE #243: fix(integrations): Circle paid-membership check fails open when active field is absent

coleam00:
Fixes #243

**Source run-id:** `b395cb43-916e-47bd-90f9-fb118a181694`

### DIFF
--- a/app/backend/integrations/circle.py
+++ b/app/backend/integrations/circle.py
@@ -64,8 +64,9 @@ async def verify_paid_member(email: str) -> bool:
             member = _extract_member(r.json())
             if member is None:
                 return False
-            if not member.get("active", True):
-                # Inactive members lose access immediately.
+            if not member.get("active", False):
+                # Inactive members — or members whose payload lacks an explicit
+                # active signal — lose access immediately (fail-closed).
                 return False
             member_id = member.get("id")
             if not member_id:
--- a/app/backend/tests/test_circle.py
+++ b/app/backend/tests/test_circle.py
@@ -110,6 +110,20 @@ async def test_inactive_member_returns_false():
 
 
 @respx.mock
+async def test_member_without_active_field_returns_false():
+    """A member payload lacking an explicit `active` key must fail closed.
+
+    Circle's search-result shape does not reliably include `active`; absence of
+    an explicit active signal must deny access on this paid-content gate.
+    """
+    respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(
+        return_value=httpx.Response(200, json={"id": 999})
+    )
+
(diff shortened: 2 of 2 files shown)


coleam00:

An exemplary minimal fix. The implementation correctly identifies that the documented fail-closed invariant demands denying access when no explicit `active` signal is present, and applies the precise one-line change (`get("active", True)` -> `get("active", False)`) that handles explicit-true, explicit-false, and absent cases correctly without the brittle identity-check alternative. Scope is tight — exactly the two files needed, with all listed out-of-scope elements (_extract_member, timeout, two-step flow) left alone — and the inline comment is upgraded to document the fail-closed reasoning. A single, well-targeted regression test proves the new behavior using the codebase's established respx mocking pattern.

The only reason it isn't a perfect 70 is that there is mild room for additional defensive coverage (e.g., asserting an explicit `active: True` still proceeds to the access-group step, or a payload with an alternate status-key shape) which would have hardened the change against the very response-shape ambiguity the issue describes. That is a minor polish point, not a defect; plan-to-implementation fidelity is essentially perfect, down to the test name.

## Scores

|---|---|---|
| solves_issue | 10/10 | Flips the default from True to False so a member payload lacking an explicit `active` key now returns False, exactly satisfying all three acceptance criteria (absent->deny, explicit-active->proceed, access-group check untouched). |
| root_cause | 10/10 | Fixes the actual defect — the `member.get("active", True)` fail-open default — rather than masking the symptom. Restores the module's documented fail-closed contract at its source. |
| scope_discipline | 10/10 | Surgical: only circle.py (one line + comment) and the test file. _extract_member, the timeout, and the two-step request flow are all left untouched per the out-of-scope list. |
| subtle_correctness | 9/10 | Correctly reasons through explicit-True (proceeds), explicit-False (denies), and absent (now denies) cases; avoids the identity-check footgun. The `get(..., False)` form is the idiomatic, type-safe one-liner. |
| code_quality | 9/10 | Clear, idiomatic change with an updated inline comment that explains the fail-closed rationale, plus a well-documented test docstring. Matches surrounding style. |
| test_discipline | 9/10 | Adds a meaningful, targeted regression test (`{"id": 999}` with no active key -> False) following the existing @respx.mock/httpx.Response pattern. Nothing gamed; existing tests preserved. Single case is sufficient for this bug. |
| plan_impl_fidelity | 10/10 | Implementation faithfully executes the plan: change True->False default and add the missing-active-key test — even the test name (`test_member_without_active_field_returns_false`) matches the plan's suggestion. No drift. |

---

PR #259, opened 2026-05-31

### LINKED ISSUE #243: fix(integrations): Circle paid-membership check fails open when active field is absent

coleam00:
Fixes #243

**Source run-id:** `2cd12dcd-17d9-43d0-bc67-ec6b7be93e2b`

### DIFF
--- a/app/backend/integrations/circle.py
+++ b/app/backend/integrations/circle.py
@@ -64,8 +64,9 @@ async def verify_paid_member(email: str) -> bool:
             member = _extract_member(r.json())
             if member is None:
                 return False
-            if not member.get("active", True):
-                # Inactive members lose access immediately.
+            if not member.get("active", False):
+                # Inactive members — or members whose status is absent/unknown —
+                # lose access immediately (fail-closed per the module contract).
                 return False
             member_id = member.get("id")
             if not member_id:
--- a/app/backend/tests/test_circle.py
+++ b/app/backend/tests/test_circle.py
@@ -110,6 +110,20 @@ async def test_inactive_member_returns_false():
 
 
 @respx.mock
+async def test_member_missing_active_field_fails_closed():
+    """A member object with no `active` field must be treated as a non-member.
+
+    Circle's search-result shape does not reliably include `active`; a
+    present-but-unknown-status member must NOT be granted access (fail-closed).
+    """
+    respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(
+        return_value=httpx.Response(200, json={"id": 999})  # no `active` key
+    )
+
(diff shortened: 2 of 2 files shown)


coleam00:

This is a tight, correct fix that nails the issue. The implementation changes the fail-open default (`active`, True -> False) so a Circle member object lacking an active field is now treated as a non-member, satisfying the fail-closed contract documented in the module while leaving explicitly-active members and the downstream access-group check intact. It respects the out-of-scope boundaries exactly (no changes to `_extract_member`, request flow, or timeout), the comment update explains the reasoning rather than just describing the line, and a targeted regression test mocks the missing-active-field response and asserts denial. Plan-to-implementation fidelity is excellent — the diff is a near-verbatim execution of the planned single-token flip plus the named regression test.

Weaknesses are minor: the fix does not proactively address the issue's hint that status may appear under a different key in search results (defensible as out of scope), and the test suite could have added an explicit positive assertion that an active member proceeds to the access-group step. Overall an excellent, surgical, well-tested change that lands at 63/70.

## Scores

|---|---|---|
| solves_issue | 9/10 | Flips `member.get("active", True)` to `False`, so a member object lacking an active field now returns False, while an explicitly active member still proceeds and the access-group check is untouched. Hits every acceptance criterion. |
| root_cause | 9/10 | Targets the actual defect — the fail-open default value — rather than masking a symptom. Correctly reasons that a fail-closed paid-access ACL must deny on absence of an explicit active signal, exactly the intended conclusion. |
| scope_discipline | 10/10 | Two-line change plus a comment in circle.py and one new test; nothing else touched. `_extract_member`, the request flow, and the timeout are all left alone per the out-of-scope list. |
| subtle_correctness | 8/10 | Edge case (absent field) handled correctly with no async/type foot-guns introduced. Does not additionally guard against the issue's hinted 'status under a different key' shape, but that is arguably out of scope and acceptance only requires the absent-field path. |
| code_quality | 9/10 | Clean, idiomatic, and the updated comment clearly documents the fail-closed rationale tied to the module contract, which aids future readers. |
| test_discipline | 8/10 | Adds a focused respx-based test asserting a member with no active key returns False, consistent with existing test style. An explicit positive test for the active-then-proceeds path would have rounded it out, but the existing inactive test plus this cover the core behavior. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim: the exact default-value flip, the optional fail-closed comment, and the named regression test `test_member_missing_active_field_fails_closed` mocking `{"id": 999}` placed near `test_inactive_member_returns_false`. No drift in either direction. |

---

PR #260, opened 2026-05-31
[bench-OK] review: remove npm package-lock.json (project uses Bun)

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

```js
  ...
}
...
...
```

### Acceptance

### Out of scope

### Why this tests reasoning

coleam00:
Fixes #244

**Source run-id:** `4850bb9d-ab4b-4ad8-b85a-a35969b4c66a`


coleam00:

This is a near-exemplary fix. It correctly diagnoses the root cause — a `break` that only escaped the inner `for` loop while the outer `while` kept draining — and applies the minimal, structurally sound remedy: labeling the outer loop, breaking out of it on the error payload, and guarding the `onComplete` call. Critically, it resists the tempting-but-wrong fixes (throwing inside the branch, calling reader.cancel(), or reordering the throw before finally), preserving the exact cleanup-once-then-throw lifecycle the issue warned about. The regression test is genuinely meaningful: it injects post-error trailing tokens and [DONE] to prove the loop stops promptly, and asserts both that onComplete is skipped and that all React streaming state resets.

The only blemish is cosmetic and outside the code itself: the PR title and self-review commit message ("remove npm package-lock.json (project uses Bun)") have nothing to do with the actual diff, which appears to be a mislabeling from the benchmark routing rather than a content problem. The implementation tracks the plan essentially perfectly. A strong, production-grade result.

## Scores

|---|---|---|
| solves_issue | 10/10 | Exactly addresses all three acceptance criteria: the labeled `break outer` stops the reader loop promptly, `onComplete` is guarded by `if (!streamError)` so it never fires on error, and the trailing `throw streamError` still surfaces the failure. |
| root_cause | 10/10 | Targets the true nested-loop control-flow defect rather than the symptom — labeling the `while` and breaking out of it is the correct structural fix, leaving the finally/throw lifecycle untouched. |
| scope_discipline | 9/10 | The diff is surgical — two files, +43/-3, touching only the buggy hook and its test, with no changes to 401/429, the SSE parser, or status handling. Docked one point because the PR title/self-review commit message reference removing a package-lock.json that does not appear in the diff (a mislabeling artifact). |
| subtle_correctness | 10/10 | Carefully preserves the finally-runs-once ordering and post-finally throw, avoids the reader.cancel()/early-return foot-guns called out in the issue, and the test deliberately appends trailing tokens + [DONE] to prove the loop no longer drains. |
| code_quality | 9/10 | Clean, minimal, idiomatic TypeScript — labeled-loop break is the right idiom and the guard reads clearly. Minor ding for the inaccurate self-review commit/PR title that doesn't describe the actual change. |
| test_discipline | 10/10 | Adds a focused regression test that asserts rejection with the right message, onComplete not called, and all streaming state reset (confirming finally ran); the trailing-token construction is a thoughtful proof that draining stopped. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan verbatim: `outer:` label, `break outer`, `if (!streamError) onComplete(...)`, untouched finally/throw, and a regression test matching the described makeSseStream/renderHook/act pattern and assertions. |

---

PR #261, opened 2026-05-31
[bench-KK] review: remove npm package-lock.json (project uses bun)

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `3ece3970-3c5a-472a-bcb9-4c974a54debe`


coleam00:

This is an excellent, textbook fix. The PR identifies and corrects the actual nested-loop control-flow defect — the inner `break` left the outer `while` spinning and draining the stream before firing onComplete with partial text — by labeling the loop (`readerLoop:`), breaking the outer loop on the error branch, and guarding `onComplete` with `if (!streamError)`. Crucially it preserves the finally-then-throw cleanup ordering the author relies on, exactly the subtle lifecycle concern the issue flagged as the trap. The accompanying regression test is targeted and meaningful, asserting onComplete is never called, the promise rejects, and React streaming state resets; it also uses the correct JSON-encoded-string SSE token format.

The only blemishes are cosmetic and non-code: the PR title and one self-review commit reference removing an npm package-lock.json, which has nothing to do with the actual diff and could confuse a reviewer scanning metadata. That did not bleed into the code — the diff stays strictly within scope. Implementation fidelity to the plan is essentially perfect, and the impl even corrected a minor error in the plan's proposed test fixture. A high-quality, minimal, root-cause fix.

## Scores

|---|---|---|
| solves_issue | 10/10 | Implements exactly the fix the issue asked for: the reader loop stops promptly on a {"error"} payload, onComplete is guarded so it does not fire, and streamError is still thrown. All three acceptance criteria are met. |
| root_cause | 10/10 | Addresses the true root cause — the inner `break` only exited the `for` loop while the outer `while` kept draining. Labeling the loop and using `break readerLoop` fixes the control-flow defect directly rather than masking the symptom. |
| scope_discipline | 9/10 | The diff is surgical: a two-line functional change in the hook plus one regression test, touching nothing in the out-of-scope 401/429, SSE parser, or status-event paths. Minor ding only because the PR title/self-review commit ('remove npm package-lock.json') is misleading and unrelated to the actual diff. |
| subtle_correctness | 10/10 | Correctly preserves the finally-then-throw lifecycle (cleanup runs exactly once, error thrown after), avoiding the foot-gun the issue warned about. The labeled break also stops draining a large stream promptly, and the test uses the correct JSON-encoded-string token format (`data: "hello"`). |
| code_quality | 9/10 | Labeled-break is idiomatic JS, minimal, and reads cleanly; the `if (!streamError)` guard is self-documenting. Production-grade for the change size. |
| test_discipline | 9/10 | Adds a focused regression test that drives tokens-then-error and asserts onComplete is not called, the promise rejects with the error message, and streaming state resets. Could additionally assert the stream stops early rather than draining, but it covers the contract well. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan faithfully — exact label, break readerLoop, and onComplete guard as specified. It even improves on the plan's test by using the correct `data: "hello"` token framing instead of the plan's incorrect `{"text":"hello"}` objects. |

---

PR #262, opened 2026-05-31

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `31cde88f-6c6a-4041-8775-0e36620f68c2`

### DIFF
--- a/app/frontend/src/hooks/useStreamingResponse.test.ts
+++ b/app/frontend/src/hooks/useStreamingResponse.test.ts
@@ -537,6 +537,55 @@ describe('conversationId reset — state clears on navigation', () => {
   });
 });
 
+describe('mid-stream error — partial text is not committed as success', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('rejects, skips onComplete, and still runs finally cleanup on a {"error"} payload', async () => {
+    // A normal token lands first, then the server emits a mid-stream error.
+    // The accumulated partial text must NOT be committed via onComplete.
+    const sseChunks = [
+      'data: "Partial answer"\n\n',
+      'data: {"error":"Mid-stream failure"}\n\n',
+      // A trailing [DONE] that some backends still emit after the error — it
+      // must never re-enable the success path once streamError is set.
+      'data: [DONE]\n\n',
+    ];
+
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        body: makeSseStream(sseChunks),
+      }),
+    );
+
+    const onComplete = vi.fn();
+    const { result } = renderHook(() => useStreamingResponse('conv-1'));
+
+    let caught: unknown;
+    await act(async () => {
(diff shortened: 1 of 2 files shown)


coleam00:

A near-textbook solution to issue #244. The implementation identifies and fixes the real root cause — nested-loop control flow — by introducing a streamAborted flag that breaks the outer while loop immediately, rather than the naive guard-only fix the issue warned against. It correctly preserves the delicate finally-then-throw ordering so React state resets still run exactly once and the caller still surfaces the error. Scope discipline is exemplary, leaving every out-of-scope concern (401/429, SSE parser, status events, finally ordering) untouched. The standout sign of careful reasoning is renaming the plan's proposed abortStream flag to streamAborted, avoiding a shadowing collision with the existing abortStream useCallback in the same hook.

The accompanying test is meaningful and targeted, exercising the partial-token-then-error sequence and even guarding against a trailing [DONE] re-enabling the success path, while asserting all three cleanup state resets. Minor, non-blocking observations: the abandoned ReadableStream reader is not explicitly cancelled on the error path (pre-existing behavior, correctly left out of scope), keeping this a clean, minimal, high-confidence fix. Overall an excellent PR with strong plan fidelity.

## Scores

|---|---|---|
| solves_issue | 10/10 | Hits all three acceptance criteria exactly: on a {"error"} payload the reader loop now breaks out of the outer while via the streamAborted flag, onComplete is guarded by if (!streamError) so partial text is never committed, streamError is still thrown after finally, and the finally cleanup runs exactly once. |
| root_cause | 10/10 | This is the true control-flow fix the issue described — it breaks the outer loop (stops draining the stream) rather than only guarding onComplete, which the issue explicitly flagged as the insufficient band-aid that races the state resets. |
| scope_discipline | 10/10 | Surgical: two files, four small edits in the hook plus one test. Leaves 401/429 handling, the SSE framing parser, and status-event handling untouched, exactly as the out-of-scope list demanded. |
| subtle_correctness | 9/10 | Thoughtful about failure modes — the test confirms a trailing [DONE] after the error cannot re-enable the success path, finally-then-throw ordering is preserved, and the flag is named streamAborted to avoid shadowing the existing abortStream useCallback. The abandoned reader isn't explicitly cancelled, but that's pre-existing and out of scope. |
| code_quality | 9/10 | Clear, well-placed comments explaining why the loop exits early and why onComplete is gated; idiomatic let-flag pattern matching the existing streamError style. Production-grade and consistent with the file. |
| test_discipline | 9/10 | Adds one focused integration test using the existing renderHook + vi.stubGlobal + makeSseStream pattern, asserting rejection with the right message, onComplete not called, and all three streaming-state resets. Even covers the trailing-[DONE] edge case. Nothing gamed. |
| plan_impl_fidelity | 9/10 | Faithfully executes the plan's two-edit-plus-test approach. The only deviations are improvements: renaming the flag from the plan's collision-prone abortStream to streamAborted, and checking the flag at the bottom of the loop rather than the top (functionally equivalent). Test uses a string error payload matching what the code actually parses. |

---

PR #263, opened 2026-05-31
[bench-OO] fix(web): stop stream drain on mid-stream error before onComplete

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `777cce92-9c9c-4da5-842b-19168656ffb6`

### DIFF
--- a/app/frontend/src/hooks/useStreamingResponse.test.ts
+++ b/app/frontend/src/hooks/useStreamingResponse.test.ts
@@ -439,6 +439,55 @@ describe('status event SSE parsing — hook state transitions', () => {
   });
 });
 
+describe('mid-stream error handling', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('stops draining and rejects without calling onComplete on a {"error"} payload', async () => {
+    // Tokens flow, then the server emits a mid-stream error. The trailing token and
+    // [DONE] must never be processed — proving the outer reader loop stopped promptly.
+    const sseChunks = [
+      'data: "Partial answer"\n\n',
+      'data: {"error": "Overloaded"}\n\n',
+      'data: " should never be appended"\n\n',
+      'data: [DONE]\n\n',
+    ];
+
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        body: makeSseStream(sseChunks),
+      }),
+    );
+
+    const onComplete = vi.fn();
+    const { result } = renderHook(() => useStreamingResponse('conv-1'));
+
+    let caught: Error | null = null;
+    await act(async () => {
+      try {
+        await result.current.startStream('conv-1', 'hi', onComplete);
+      } catch (e) {
+        caught = e as Error;
+      }
+    });
+
+    // streamError is surfaced to the caller...
+    expect(caught).toBeInstanceOf(Error);
+    expect(caught?.message).toBe('Overloaded');
+    // ...and the partial answer was never committed as a success.
+    expect(onComplete).not.toHaveBeenCalled();
+    // The finally block still reset all streaming state exactly once.
+    expect(result.current.isStreaming).toBe(false);
+    expect(result.current.streamingContent).toBe('');
+    expect(result.current.streamingSources).toHaveLength(0);
+    expect(result.current.streamingStatus).toBeNull();
+  });
+});
+
 describe('abortStream', () => {
   beforeEach(() => {
     vi.clearAllMocks();
--- a/app/frontend/src/hooks/useStreamingResponse.ts
+++ b/app/frontend/src/hooks/useStreamingResponse.ts
@@ -108,6 +108,9 @@ export function useStreamingResponse(conversationId: string | null) {
         let buffer = '';
 
         while (true) {
+          // A mid-stream {"error"} payload only breaks the inner for-loop below; stop
+          // draining the reader promptly here so onComplete never fires with partial text.
+          if (streamError) break;
           const { done, value } = await reader.read();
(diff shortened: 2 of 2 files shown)


coleam00:

This is a near-exemplary fix. It correctly diagnoses the nested-loop control-flow bug — that the inner break only exits the for-loop while the outer while keeps draining and then unconditionally calls onComplete with truncated text — and resolves it with the minimal, structurally-sound approach: a streamError flag check at the top of the while loop plus an if(!streamError) guard on onComplete, leaving the finally cleanup and trailing throw exactly as the author intended. Scope discipline is excellent, the regression test is purpose-built to prove the outer loop stops (trailing token + [DONE] must never be processed), and the change reads as if written by the original author.

The only quibble is the placement of the guard at the top of the loop rather than immediately after the inner break, meaning the remainder of the current chunk's inner-for iteration finishes before the outer loop exits; this is harmless because streamError is already set by the error token and onComplete is guarded, so no partial text is ever committed. Plan-to-implementation fidelity is total — the diff is effectively the plan's pseudocode realized line-for-line, including the test fixture. A strong OO (Opus/Opus) cell result.

## Scores

|---|---|---|
| solves_issue | 10/10 | Exactly implements all three acceptance criteria: the reader loop stops promptly on a {"error"} payload via a top-of-while guard, onComplete is guarded with if(!streamError) so it never fires with partial text, and streamError is still thrown after the unchanged finally block. |
| root_cause | 10/10 | Addresses the true root cause — the nested-loop control-flow defect where the inner break only exited the for-loop. The flag-at-top-of-while check propagates the break to the outer loop, exactly the structural fix the issue called for, not a symptom patch. |
| scope_discipline | 10/10 | Surgical: two files, +56/-2, only the one production hook and its test. Explicitly left the 401/429 handling, SSE parser, and status-event branches untouched as required by the out-of-scope section. |
| subtle_correctness | 9/10 | Handles the lifecycle correctly — finally still runs exactly once, throw ordering preserved, no double-cleanup. Chose the flag check over a labeled break to avoid lint friction. Minor nit: the guard sits at the top so one extra inner-loop iteration completes before the break, but that's after the error token already set streamError, so no partial text leaks. |
| code_quality | 9/10 | Clean, idiomatic, with clear comments explaining why each guard exists. Matches the surrounding style. Readable and production-grade. |
| test_discipline | 10/10 | Adds a meaningful targeted regression test in a new describe block, using the established makeSseStream/stubGlobal/renderHook pattern. Includes a trailing token + [DONE] after the error specifically to prove draining stopped, and asserts rejection, onComplete-not-called, and all state reset. Existing tests left intact. |
| plan_impl_fidelity | 10/10 | The implementation faithfully executes the plan verbatim — the flag-at-top-of-while guard, the if(!streamError) onComplete guard, the untouched finally/throw, and the exact test scaffold (including the identical sseChunks array) all match the plan's specific changes. |

---

coleam00:

An excellent, textbook-quality fix. It correctly diagnoses the nested-loop control-flow bug — where the inner `break` on a mid-stream `{"error"}` payload fails to stop the outer `while` reader loop — and resolves it with a minimal flag check at the top of the loop plus an `if (!streamError)` guard on onComplete. Crucially, it preserves the intentional cleanup lifecycle: the finally block still runs exactly once and the error is thrown afterward, avoiding the trap of throwing inside the error branch that the issue explicitly warned against. Scope is tightly disciplined to the two intended files, and the regression test is genuinely meaningful, using a post-error trailing token plus [DONE] that must never be processed to prove the drain actually stops rather than merely suppressing onComplete.

The implementation tracks the planning artifact almost line-for-line, indicating tight plan/impl coordination. The only minor reservations are cosmetic: the flag-check approach still finishes processing the current already-buffered batch of events before exiting (acceptable and arguably correct), and subtle-correctness loses a point only for that nuance. Overall this is among the strongest possible responses to the issue — root-cause-correct, surgical, idiomatic, and well-tested.

## Scores

|---|---|---|
| solves_issue | 10/10 | Exactly addresses the issue: the top-of-while flag check stops the reader drain promptly, onComplete is guarded by `if (!streamError)`, and the post-finally throw is preserved so the caller still surfaces the failure. All three acceptance criteria are met. |
| root_cause | 10/10 | Fixes the actual nested-loop control-flow defect (inner `break` not propagating to the outer `while`) rather than papering over the symptom. It correctly avoids the naive throw-in-branch that would have disturbed the finally cleanup ordering. |
| scope_discipline | 10/10 | Surgical: only the two intended files touched (hook + its test), three small lines added to production code. No drift into 401/429, SSE framing, or status handling, exactly per the out-of-scope list. |
| subtle_correctness | 9/10 | Correctly reasons about the loop/cleanup lifecycle: inner break sets streamError, outer guard exits before the next read, finally runs exactly once, then the error throws. Edge cases handled; minor deduction only because the flag-check leaves one already-buffered event partially processed, which is acceptable and matches intent. |
| code_quality | 9/10 | Clear, well-placed comments explaining the why, idiomatic flag-check approach that sidesteps labeled-break lint friction, consistent with the surrounding code style. Production-grade. |
| test_discipline | 9/10 | Strong targeted regression test using the established makeSseStream/stubGlobal/renderHook pattern, with a trailing token after the error to actively prove the drain stops; asserts onComplete not called, error surfaced, and all state reset once. Existing happy-path tests left intact. |
| plan_impl_fidelity | 10/10 | The implementation is a near-verbatim execution of source-plan.md: the top-of-loop flag check, the `if (!streamError)` onComplete guard, untouched finally/throw, and the exact SSE-chunk regression test the plan specified. No drift. |

---

PR #264, opened 2026-05-31
[bench-OO] fix(rag): guard timestamp distribution for zero-duration segments

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

```python
if len(sub_chunks) > 1:
    duration = end_s - start_s
```

### Acceptance

### Out of scope

### Why this tests reasoning

coleam00:
Fixes #245

**Source run-id:** `e69d91df-8d99-420f-a043-b694ffd6c6df`

### DIFF
--- a/app/backend/rag/chunker.py
+++ b/app/backend/rag/chunker.py
@@ -162,9 +162,12 @@ def chunk_video_timestamped(segments: list[TimestampedSegment]) -> tuple[list[di
 
             # Distribute timestamps evenly across sub-chunks when HybridChunker
             # splits a segment into multiple pieces (no worse than fallback's
-            # proportional timestamps, which are explicitly accepted).
-            if len(sub_chunks) > 1:
-                duration = end_s - start_s
+            # proportional timestamps, which are explicitly accepted). When the
+            # segment has zero or negative duration (e.g. the final Supadata
+            # segment with no explicit end), even distribution is meaningless, so
+            # leave each sub-chunk on the original [start_s, end_s] boundary.
+            duration = end_s - start_s
+            if len(sub_chunks) > 1 and duration > 0:
                 step = duration / len(sub_chunks)
                 for i, sc in enumerate(sub_chunks):
                     sc["start_seconds"] = start_s + i * step
--- a/app/backend/tests/test_chunker_timestamps.py
+++ b/app/backend/tests/test_chunker_timestamps.py
@@ -8,7 +8,18 @@ Verifies:
   - chunk_video_fallback evenly distributes estimated duration across chunks
 """
 
+import pytest
+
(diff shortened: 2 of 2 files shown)


coleam00:

An excellent, textbook fix. The change correctly identifies that even-distribution is only meaningful for positive-duration segments and guards the branch accordingly (`duration > 0`), which simultaneously handles the zero-duration final-Supadata-segment case from the issue and the negative-duration malformed case, without inventing fake durations via step clamping — the trap the issue explicitly warned against. Scope is tight (4 lines of logic plus tests), the comment clearly documents intent, and the change respects every out-of-scope boundary (_parse_segments, fallback path, snippet/contextualize all untouched).

Test discipline is strong: the zero-, negative-, and positive-duration cases each assert the segment actually split (len > 1) so the guarded branch is provably exercised, preventing a vacuously-passing test. Plan-to-impl fidelity is near-perfect — the diff mirrors the plan almost verbatim. The only nitpick is cosmetic: zero-duration sub-chunks still all share start==end==start_s, but that is precisely the preserved-boundary behavior the issue accepted as no-worse-than-fallback, not a regression. Overall this is production-grade work.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly fixes the degenerate collapse by gating the even-distribution branch on `duration > 0`; zero/negative-duration splittable segments keep their original boundary, normal segments still distribute, and chunk_video_fallback is untouched — every acceptance criterion is met. |
| root_cause | 10/10 | Targets the true root cause: when duration <= 0 the even distribution is meaningless, so it skips redistribution rather than clamping step or inventing fake durations — exactly the reasoning the issue called out as correct. |
| scope_discipline | 10/10 | Surgical: a 4-line change in chunker.py plus a test file. No touches to _parse_segments, snippet/contextualize logic, or the fallback path, all of which were declared out of scope. |
| subtle_correctness | 9/10 | Correctly uses `duration > 0` so it also handles the negative-duration (end < start) case, avoiding decreasing spans; computing duration before the branch is clean and the single-sub-chunk path is unaffected. Minor note only: zero-duration sub-chunks still carry start==end (the accepted preserved-boundary behavior), not a defect. |
| code_quality | 10/10 | Clear, well-worded comment explaining the why, idiomatic to the surrounding code, no churn beyond the necessary lines. |
| test_discipline | 10/10 | Three meaningful, well-named tests (zero-duration, negative-duration, positive control), each asserting branch reachability (len > 1) so they can't silently pass, with pytest.approx for the float distribution — genuine regression coverage, no gaming. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan faithfully down to the comment text, the `duration = end_s - start_s` hoist, the `duration > 0` guard, and the exact trio of tests (zero/negative/positive) specified in the plan's Tests section. |

---

PR #265, opened 2026-05-31

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `7e8aaa45-3223-4f58-b73d-6ccf9177e83f`

### DIFF
--- a/app/backend/rag/chunker.py
+++ b/app/backend/rag/chunker.py
@@ -163,8 +163,8 @@ def chunk_video_timestamped(segments: list[TimestampedSegment]) -> tuple[list[di
             # Distribute timestamps evenly across sub-chunks when HybridChunker
             # splits a segment into multiple pieces (no worse than fallback's
             # proportional timestamps, which are explicitly accepted).
-            if len(sub_chunks) > 1:
-                duration = end_s - start_s
+            duration = end_s - start_s
+            if len(sub_chunks) > 1 and duration > 0:
                 step = duration / len(sub_chunks)
                 for i, sc in enumerate(sub_chunks):
                     sc["start_seconds"] = start_s + i * step
--- a/app/backend/tests/test_chunker_timestamps.py
+++ b/app/backend/tests/test_chunker_timestamps.py
@@ -39,6 +39,28 @@ class TestChunkVideoTimestamped:
         # The snippet should be the raw segment text (up to 300 chars)
         assert chunk["snippet"] == "Original transcript text here"
 
+    def test_zero_duration_segment_keeps_boundary_when_split(self) -> None:
+        """A zero-duration segment that splits into multiple sub-chunks must not
+        collapse every sub-chunk onto a single instant.
+
+        Regression for issue #245: the final transcript segment can have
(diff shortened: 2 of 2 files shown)


coleam00:

An excellent, textbook fix. The change identifies and addresses the true root cause described in the issue — when a segment has duration <= 0 the even-distribution is meaningless, so the code falls through and keeps each sub-chunk on the original [start_s, end_s] boundary, only redistributing when duration > 0. It sidesteps both tempting wrong fixes (clamping step to a minimum, or special-casing duration == 0 while still pinning to start_s). Scope is surgical: a 2-line edit in chunker.py plus a single targeted regression test, with no collateral changes to _parse_segments or the fallback path.

Weaknesses are minor. The regression test depends on HybridChunker genuinely splitting the synthetic 200-sentence text into multiple sub-chunks (a mild brittleness), and it does not add an explicit positive-duration assertion, instead trusting the existing suite to guard the even-distribution behavior. Neither materially undermines the fix. Implementation fidelity to the plan is complete. Overall a high-quality, well-reasoned PR.

## Scores

|---|---|---|
| solves_issue | 10/10 | Adds the `duration > 0` guard so zero-/negative-duration segments stop collapsing every sub-chunk onto start_s, while positive durations keep even distribution and the fallback path is untouched — every acceptance criterion is met exactly. |
| root_cause | 10/10 | Targets the real root cause: when duration <= 0 the even-distribution math is meaningless, so it falls through and keeps the original [start_s, end_s] boundary rather than clamping step or faking a duration. This is exactly the correct reasoning the issue called out and avoids the tempting band-aids. |
| scope_discipline | 10/10 | Surgical 2-line change in chunker.py plus one new test (24 additions, 2 deletions across 2 files). No unrelated edits, no touching _parse_segments or the fallback path. |
| subtle_correctness | 9/10 | Correctly relies on the earlier per-sub-chunk assignment of start_s/end_s so falling through preserves the boundary, and the `> 0` check also handles negative durations. Sound; only nuance is that a genuinely zero-width span still renders for end==start segments, which is inherent to the source data and out of scope. |
| code_quality | 9/10 | Minimal, idiomatic edit that hoists `duration` and extends the existing condition; reads naturally with the surrounding comment. Production-grade for a fix this size. |
| test_discipline | 9/10 | Adds a focused regression test that constructs a zero-duration segment with text long enough to split into >1 sub-chunk and asserts all boundaries stay at 330.0. Meaningful and not gamed; slightly brittle in depending on HybridChunker actually splitting, and positive-duration redistribution relies on pre-existing tests rather than a new assertion. |
| plan_impl_fidelity | 10/10 | The implementation faithfully executes the plan: the `len(sub_chunks) > 1 and duration > 0` guard and the regression test with start=end=330.0 asserting len>1 and equal boundaries match the plan's specified changes verbatim. |

---

PR #266, opened 2026-05-31

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `1547b56d-67f3-4d63-be24-a4052bee0d72`

### DIFF
--- a/app/backend/rag/chunker.py
+++ b/app/backend/rag/chunker.py
@@ -165,10 +165,11 @@ def chunk_video_timestamped(segments: list[TimestampedSegment]) -> tuple[list[di
             # proportional timestamps, which are explicitly accepted).
             if len(sub_chunks) > 1:
                 duration = end_s - start_s
-                step = duration / len(sub_chunks)
-                for i, sc in enumerate(sub_chunks):
-                    sc["start_seconds"] = start_s + i * step
-                    sc["end_seconds"] = start_s + (i + 1) * step
+                if duration > 0:
+                    step = duration / len(sub_chunks)
+                    for i, sc in enumerate(sub_chunks):
+                        sc["start_seconds"] = start_s + i * step
+                        sc["end_seconds"] = start_s + (i + 1) * step
 
             results.extend(sub_chunks)
         except Exception as exc:
--- a/app/backend/tests/test_chunker_timestamps.py
+++ b/app/backend/tests/test_chunker_timestamps.py
@@ -55,6 +55,36 @@ class TestChunkVideoTimestamped:
         # Should not produce any chunks from the empty segment
         assert all("Real content" in c["content"] or "Real content" in c["snippet"] for c in result)
 
+    def test_zero_duration_segment_preserves_boundary(self) -> None:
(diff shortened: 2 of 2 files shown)


coleam00:

An exemplary, tightly-scoped fix. The KK cell (plan=Kimi, impl=Kimi) correctly diagnosed the two-file correctness property the issue was probing: that the even-distribution branch is meaningless when a zero-width final segment (produced by _parse_segments' float(start) fallback and Supadata's zero-duration tail) feeds in, and that the right move is to keep each sub-chunk on the original [start_s, end_s] boundary rather than fabricating a minimum step. The plan articulated this reasoning explicitly and the implementation matched it line-for-line, touching only chunker.py and its test file while leaving _parse_segments and chunk_video_fallback untouched per the out-of-scope constraints.

Weaknesses are minor and cosmetic: the comment above the guarded block still describes only the even-distribution rationale and wasn't extended to note the duration <= 0 branch, and the test suite, while strong, could have added an explicit negative-duration assertion alongside the zero-duration one. Neither affects correctness. This is production-grade work that avoids both tempting wrong fixes the issue warned about.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly satisfies all three acceptance criteria: zero-duration (end_s <= start_s) segments splitting into >1 sub-chunks no longer collapse to all-equal timestamps, positive-duration even-distribution is preserved, and chunk_video_fallback is untouched. |
| root_cause | 10/10 | Implements exactly the correct reasoning the issue called for — guard the distribution with `duration > 0` so each sub-chunk retains its original [start_s, end_s] boundary, rather than clamping step or special-casing while still pinning to start_s. No fake durations invented. |
| scope_discipline | 10/10 | Surgical: 35 additions / 4 deletions across exactly the two files the issue and plan named, the chunker change is a 1-level indent + guard, and _parse_segments / fallback / snippet logic are left alone. |
| subtle_correctness | 9/10 | Correctly relies on sub_chunks being pre-initialized with start_s/end_s so skipping the loop yields a sane span; `> 0` cleanly covers both end_s == start_s and the negative-duration case. No type or foot-gun regressions; minor nit that the inline comment block above wasn't updated to mention the new branch. |
| code_quality | 9/10 | Readable and idiomatic, matches surrounding style; the guard reads naturally. Slight room for a one-line comment explaining why duration <= 0 preserves the original boundary, but the intent is clear from context. |
| test_discipline | 9/10 | Two meaningful, targeted regression tests: one asserts every sub-chunk keeps 120.0/120.0 for a zero-duration split, the other verifies positive-duration distribution stays monotonic and evenly spaced (0->100). Real assertions, nothing gamed; could add an explicit negative-duration case but coverage is solid. |
| plan_impl_fidelity | 10/10 | source-plan.md is present and the diff executes it faithfully — the exact `if duration > 0:` guard at the named location plus the zero-duration regression test, and it also added the positive-duration test the plan flagged as conditional. No drift. |

---

PR #267, opened 2026-05-31

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `a85dce75-8b20-4be9-8cb0-7ce3b5f83c1c`

### DIFF
--- a/app/backend/rag/chunker.py
+++ b/app/backend/rag/chunker.py
@@ -165,10 +165,14 @@ def chunk_video_timestamped(segments: list[TimestampedSegment]) -> tuple[list[di
             # proportional timestamps, which are explicitly accepted).
             if len(sub_chunks) > 1:
                 duration = end_s - start_s
-                step = duration / len(sub_chunks)
-                for i, sc in enumerate(sub_chunks):
-                    sc["start_seconds"] = start_s + i * step
-                    sc["end_seconds"] = start_s + (i + 1) * step
+                if duration > 0:
+                    step = duration / len(sub_chunks)
+                    for i, sc in enumerate(sub_chunks):
+                        sc["start_seconds"] = start_s + i * step
+                        sc["end_seconds"] = start_s + (i + 1) * step
+                # else: zero-width/inverted segment (e.g. final _parse_segments
+                # segment where end == start) — even distribution is meaningless,
+                # so each sub-chunk keeps the original [start_s, end_s] boundary.
 
             results.extend(sub_chunks)
         except Exception as exc:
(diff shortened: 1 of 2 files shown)


coleam00:

A tight, correct, well-scoped fix that nails the intended reasoning: it guards the even-distribution branch on duration > 0 and otherwise leaves each sub-chunk on its original [start_s, end_s] boundary, sidestepping the tempting clamp/fabricate anti-fix the issue warned about. The guard correctly covers both zero-width and inverted segments, the diff is minimal and idiomatic with a helpful comment, scope discipline is perfect, and the implementation tracks the Opus plan almost verbatim.

The one real weakness is test quality. Because each sub-chunk is already constructed with [start_s, end_s] and the old code's step=0 also produced all-[100,100] for an end==start segment, the flagship regression test would pass on the unpatched code — it doesn't actually prove the fix. The behavioral difference only appears for inverted (end < start) segments, which the tests never exercise. The positive-duration test partially redeems the suite by locking in the unchanged normal path. Net: an excellent fix with a regression test that doesn't quite earn its name.

## Scores

|---|---|---|
| solves_issue | 8/10 | Implements exactly the issue's prescribed correct reasoning: guard distribution on duration > 0 and otherwise keep each sub-chunk on the original [start_s, end_s] boundary, avoiding the step=0 collapse and the explicitly-forbidden clamp/fabricate fix. Note the end==start case inherently still renders a zero-width span (unfixable without fabricating data, which is out of scope), so the literal 'no all-equal timestamps' acceptance bullet is only truly satisfied for inverted segments. |
| root_cause | 9/10 | Structurally sound: it removes the meaningless even-distribution when duration <= 0 rather than papering over the symptom, correctly connecting the chunker branch to how _parse_segments/Supadata produce a zero/negative-width final segment. |
| scope_discipline | 10/10 | Surgical — only the redistribution block in chunker.py plus two regression tests. _parse_segments, the fallback path, and snippet/contextualize logic are all untouched, exactly per the out-of-scope list. |
| subtle_correctness | 9/10 | Uses `duration > 0`, so both zero-width and inverted (end < start) segments are correctly guarded, not just the == case. No type/async foot-guns introduced and the positive-duration path is byte-for-byte identical. |
| code_quality | 9/10 | Clean, idiomatic, minimal diff with a clear explanatory comment that matches the surrounding style and explains the why. Comment is correctly placed at the else level (improving on the plan's slightly misindented snippet). |
| test_discipline | 6/10 | The positive-duration test is genuinely useful and guards the unchanged-behavior acceptance. But the headline zero-duration test uses end==start, where the OLD buggy code also yields step=0 -> all [100,100]; it therefore passes on unpatched code and is non-distinguishing. An inverted segment (end < start) asserting no chunk has end < start would have been the true regression guard. |
| plan_impl_fidelity | 10/10 | The implementation faithfully executes the plan: the duration>0 guard, the comment text, the two named tests (zero-duration + positive-duration), and the out-of-scope boundaries all match the plan almost verbatim, even correcting the plan snippet's comment indentation. |

---

PR #268, opened 2026-05-31
[bench-OO] fix(rag): send extended-tier cache_control.ttl as Anthropic '1h' duration

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

```python
if tier == "extended":
```

### Acceptance

### Out of scope

### Why this tests reasoning

coleam00:
Fixes #246

**Source run-id:** `4462a806-0394-48e9-a712-a173aa7a1435`

### DIFF
--- a/app/backend/config.py
+++ b/app/backend/config.py
@@ -161,8 +161,6 @@ CATALOG_ENABLED: bool = os.environ.get("CATALOG_ENABLED", "false").strip().lower
 CATALOG_TIER: str = (
     os.environ.get("CATALOG_TIER", "standard").strip().lower()
 )  # "standard" or "extended"
-# TTL in seconds for the extended prompt-cache tier (Anthropic API requires an integer).
-CATALOG_CACHE_TTL_SECONDS: int = int(os.environ.get("CATALOG_CACHE_TTL_SECONDS", "3600"))
 
 # Cap on how many characters the get_video_transcript tool returns to the
 # model. Long videos can produce 40K+ tokens of transcript; beyond ~30K
--- a/app/backend/rag/catalog.py
+++ b/app/backend/rag/catalog.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import logging
 
-from backend.config import CATALOG_CACHE_TTL_SECONDS
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
@@ -78,7 +77,7 @@ def build_catalog_block(videos: list[dict], tier: str) -> dict:
 
     cache_control: dict = {"type": "ephemeral"}
     if tier == "extended":
-        cache_control["ttl"] = CATALOG_CACHE_TTL_SECONDS  # integer seconds per Anthropic API
+        cache_control["ttl"] = "1h"  # Anthropic cache_control.ttl enum: "5m" or "1h"
 
     return {
         "type": "text",
--- a/app/backend/tests/test_catalog.py
+++ b/app/backend/tests/test_catalog.py
@@ -71,7 +71,7 @@ def test_build_catalog_block_standard() -> None:
 def test_build_catalog_block_extended() -> None:
     videos = [{"id": "vid-abc", "title": "My Video", "url": "https://youtube.com/watch?v=abc"}]
     block = catalog.build_catalog_block(videos, "extended")
-    assert block["cache_control"] == {"type": "ephemeral", "ttl": 3600}
+    assert block["cache_control"] == {"type": "ephemeral", "ttl": "1h"}
 
 
 def test_build_catalog_block_untitled_fallback() -> None:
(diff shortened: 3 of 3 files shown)


coleam00:

An exemplary small fix. It correctly diagnoses that the bug is not just a bad literal but a wrong abstraction — the config's integer-seconds TTL — and removes that abstraction rather than papering over it, mapping the tier directly to Anthropic's "1h" duration enum and sidestepping the tempting "3600" stringify trap. Scope is surgical (fix + the dead constant/import it required + the two affected tests), and the tests are genuinely re-pointed at the correct contract rather than merely flipped to pass, with the previously misnamed integer-assertion test renamed and rewritten. Implementation fidelity to the plan is essentially perfect.

Weaknesses are negligible: env entries for the removed CATALOG_CACHE_TTL_SECONDS variable likely still linger in deploy compose as a harmless no-op, but touching deploy/ would have been out of scope for a non-deployment issue, so leaving them is the defensible call. There is no behavioral edge case left unhandled. This is production-grade work for the size of the change.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control.ttl == "1h", standard tier stays plain {"type": "ephemeral"}, and catalog block content is untouched — every acceptance criterion is met exactly. |
| root_cause | 10/10 | Correctly recognizes the config's 'TTL in seconds' abstraction is itself wrong, removes the CATALOG_CACHE_TTL_SECONDS constant entirely, and maps the tier to Anthropic's duration enum rather than stringifying the int — the true structural fix the issue framed as the hard path. |
| scope_discipline | 10/10 | Three files, all strictly necessary: the fix, the dead config constant/import it depended on, and the two tests asserting the old contract. No unrelated churn. |
| subtle_correctness | 9/10 | Avoids the tempting "3600" stringify foot-gun and uses the correct enum value. Only minor residue: deploy compose env entries for the removed var become a harmless no-op, deliberately and correctly left untouched per CLAUDE.md deploy rules. |
| code_quality | 9/10 | Idiomatic and readable; the replacement comment accurately documents the Anthropic enum ("5m" or "1h") instead of perpetuating the wrong belief. Clean removal of the now-unused import. |
| test_discipline | 9/10 | Tests are meaningfully updated rather than gamed: the misleadingly-named ttl_is_integer test is renamed to ttl_is_duration_string and rewritten to assert the correct contract (=="1h" and isinstance str), and the equality test is updated to the new expected dict. |
| plan_impl_fidelity | 10/10 | The implementation executes the source plan verbatim — same five edits across catalog.py, config.py, and test_catalog.py, including the test rename and the decision to leave deploy/ env entries alone. No drift. |

---

coleam00:

An exemplary small fix. It addresses the true root cause rather than the symptom: recognizing that Anthropic's cache_control.ttl is a duration enum ("5m"/"1h") and that the local 'TTL in seconds' config abstraction was fundamentally wrong, it maps the extended tier directly to "1h" and removes the dead CATALOG_CACHE_TTL_SECONDS constant, its incorrect comment, and the now-unused import. It sidesteps the documented trap of stringifying 3600 and preserves standard-tier behavior and the catalog block content exactly. Tests are corrected to assert the right contract, including renaming a test whose name encoded the wrong belief.

Weaknesses are negligible. Removing the config constant means an operator who had set CATALOG_CACHE_TTL_SECONDS in their env would now have it silently ignored, but that variable was already broken so the practical risk is nil, and the exploration explicitly weighed this. The implementation is a faithful, surgical execution of the plan with strong scope discipline.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control.ttl = "1h" (a valid Anthropic duration), standard tier remains plain {"type": "ephemeral"} with no ttl, and the catalog block text/inclusion is untouched. Hits every acceptance bullet exactly. |
| root_cause | 10/10 | Treats the bug at its source: the 'TTL in seconds' abstraction was itself wrong, so the fix maps tier->enum and deletes the misleading CATALOG_CACHE_TTL_SECONDS constant, its incorrect comment, and the unused import. Not a stringify-the-int band-aid. |
| scope_discipline | 9/10 | Three files, all necessary (catalog logic, dead config constant, two tests). Removing the now-dead config constant is mildly beyond a strict one-line fix but is explicitly justified by the issue (the seconds abstraction is wrong) and avoids leaving misleading dead config. |
| subtle_correctness | 9/10 | Correctly avoids the tempting "3600" string trap, cleans up the import to prevent a ruff F401, and the comment now documents the real enum ("5m"/"1h"). A stray CATALOG_CACHE_TTL_SECONDS env in deploy/ becomes a harmless no-op, correctly left alone. |
| code_quality | 9/10 | Minimal, readable, idiomatic. The replacement comment accurately documents the Anthropic contract instead of restating the old wrong belief. No naming or style issues. |
| test_discipline | 9/10 | Both assertions updated meaningfully; the misleadingly-named test_..._ttl_is_integer is renamed to ..._is_duration_string with body rewritten to assert == "1h" and isinstance str. Tests were corrected to encode the right contract, not gamed to pass. |
| plan_impl_fidelity | 10/10 | The diff executes the plan almost verbatim: same catalog.py edit and comment text, same config.py constant+comment+import removal, same two test rewrites including the exact rename and body. No drift. |

---

PR #269, opened 2026-05-31

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `d298865f-e7e9-4449-8db0-4f7636a8799c`

### DIFF
--- a/app/backend/rag/catalog.py
+++ b/app/backend/rag/catalog.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import logging
 
-from backend.config import CATALOG_CACHE_TTL_SECONDS
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
@@ -78,7 +77,7 @@ def build_catalog_block(videos: list[dict], tier: str) -> dict:
 
     cache_control: dict = {"type": "ephemeral"}
     if tier == "extended":
-        cache_control["ttl"] = CATALOG_CACHE_TTL_SECONDS  # integer seconds per Anthropic API
+        cache_control["ttl"] = "1h"  # Anthropic accepts only "5m" or "1h"
 
     return {
         "type": "text",
--- a/app/backend/tests/test_catalog.py
+++ b/app/backend/tests/test_catalog.py
@@ -71,7 +71,7 @@ def test_build_catalog_block_standard() -> None:
 def test_build_catalog_block_extended() -> None:
     videos = [{"id": "vid-abc", "title": "My Video", "url": "https://youtube.com/watch?v=abc"}]
     block = catalog.build_catalog_block(videos, "extended")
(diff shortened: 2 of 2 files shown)


coleam00:

An exemplary small-scope reasoning fix. The PR correctly identifies that Anthropic's cache_control.ttl accepts only the duration enums "5m"/"1h" and that the local 'integer seconds' abstraction (and its comment) was simply wrong. It sidesteps the tempting-but-wrong stringify-the-int trap, removes the now-unused CATALOG_CACHE_TTL_SECONDS import from catalog.py while correctly leaving the config constant alone per the issue's out-of-scope note, and corrects the inline comment. Scope is surgical (2 files, +8/-10) and the implementation is a faithful, verbatim execution of the plan.

The only minor knocks are inherent to the task's small surface area: a misleading-but-out-of-scope orphaned constant remains in config.py, and the test suite stops at unit-level assertions on the dict shape rather than verifying the value against an authoritative source. Neither is a defect given the issue's explicit boundaries. The previously bug-encoding test was honestly corrected and renamed rather than deleted or gamed. Strong across all seven dimensions.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends the valid Anthropic duration string "1h"; standard tier still emits plain {"type": "ephemeral"} with no ttl; catalog block text and inclusion logic untouched. Hits every acceptance criterion exactly. |
| root_cause | 9/10 | Correctly recognizes the 'TTL in seconds' abstraction is itself wrong and maps the tier to Anthropic's enum, dropping the now-meaningless CATALOG_CACHE_TTL_SECONDS import rather than stringifying it. True root-cause fix; leaving the orphaned config constant is per the issue's out-of-scope, not a band-aid. |
| scope_discipline | 10/10 | Surgical: only catalog.py and its test file, +8/-10 lines. No unrelated edits, no config churn. |
| subtle_correctness | 9/10 | Avoids the tempting wrong fix of stringifying 3600 to "3600"; uses the actual API-accepted enum "1h". Updated inline comment is now accurate. No type or async foot-guns introduced. |
| code_quality | 9/10 | Readable and idiomatic; removed the unused import, replaced the misleading 'integer seconds per Anthropic API' comment with a correct one. Production-grade for a one-line behavior change. |
| test_discipline | 9/10 | Both affected tests updated to assert the correct "1h" string, and the misnamed test_..._ttl_is_integer was properly renamed to _is_string and rewritten to assert str. This corrects a test that previously encoded the bug rather than gaming it to pass; standard-tier test left untouched. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan verbatim — removed the import, set ttl="1h", updated test_build_catalog_block_extended, and renamed the integer test to _is_string with str assertions. Zero drift. |

---

PR #270, opened 2026-05-31

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `b426d160-504f-427f-b688-99276cef3e59`

### DIFF
--- a/app/backend/config.py
+++ b/app/backend/config.py
@@ -161,8 +161,6 @@ CATALOG_ENABLED: bool = os.environ.get("CATALOG_ENABLED", "false").strip().lower
 CATALOG_TIER: str = (
     os.environ.get("CATALOG_TIER", "standard").strip().lower()
 )  # "standard" or "extended"
-# TTL in seconds for the extended prompt-cache tier (Anthropic API requires an integer).
-CATALOG_CACHE_TTL_SECONDS: int = int(os.environ.get("CATALOG_CACHE_TTL_SECONDS", "3600"))
 
 # Cap on how many characters the get_video_transcript tool returns to the
 # model. Long videos can produce 40K+ tokens of transcript; beyond ~30K
--- a/app/backend/rag/catalog.py
+++ b/app/backend/rag/catalog.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import logging
 
-from backend.config import CATALOG_CACHE_TTL_SECONDS
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


coleam00:

A textbook-correct fix for a reasoning-trap issue. It identifies the real root cause — the 'TTL in seconds' abstraction is fundamentally wrong because Anthropic's cache_control.ttl only accepts duration enums — and deletes the misleading constant rather than stringifying the integer (the tempting wrong fix the issue explicitly calls out). The change is surgical (3 files), the standard-tier path is correctly left untouched, and the tests are updated honestly: the integer-asserting test is renamed and rewritten to verify the string "1h", and the now-obsolete patch of the deleted constant is removed. Implementation matches the plan exactly. The only blemish is an incomplete cleanup outside the diff's lane: deploy/docker-compose.yml still passes CATALOG_CACHE_TTL_SECONDS to both blue and green app services, which is now dead config since config.py no longer reads it. This is harmless at runtime and arguably out-of-scope given the repo's deploy-modification rules, but it leaves a slightly misleading dangling reference. Overall an excellent, high-confidence PR.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control.ttl = "1h" (the valid Anthropic duration enum), standard tier still emits plain {"type": "ephemeral"} with no ttl, and the catalog block text/inclusion is unchanged. Exactly the stated acceptance criteria. |
| root_cause | 9/10 | Correctly recognizes that the 'TTL in seconds' abstraction is itself wrong and deletes the CATALOG_CACHE_TTL_SECONDS constant rather than papering over it by stringifying 3600. True root-cause fix; only minor gap is the orphaned docker-compose env var left behind. |
| scope_discipline | 9/10 | Surgical: three files (catalog.py, config.py, test_catalog.py), 8 additions / 12 deletions, nothing unrelated touched. Leaving deploy/docker-compose.yml's now-dead env reference is defensible under the repo's don't-touch-deploy rule, costing a point only for the slightly stale config. |
| subtle_correctness | 8/10 | Avoids the tempting wrong fix (the "3600" string trap) and uses the correct API enum value, demonstrating knowledge of the Anthropic prompt-cache contract over the misleading local comment. Minor foot-gun: docker-compose still injects CATALOG_CACHE_TTL_SECONDS which is now read by nothing. |
| code_quality | 9/10 | Clean, idiomatic diff: dead import and dead constant removed, the wrong comment deleted, direct "1h" assignment is self-documenting. Reads naturally in the surrounding style. |
| test_discipline | 9/10 | Both affected tests updated meaningfully, not gamed: the integer-asserting test is renamed (test_..._ttl_is_one_hour) and now asserts both type str and value "1h", and the obsolete @patch of the deleted constant is removed. Targeted and honest. |
| plan_impl_fidelity | 10/10 | The diff executes the plan item-for-item: remove the import and constant, map extended -> "1h", update test_build_catalog_block_extended to "1h", rename and rewrite the integer test, and drop the patch. No drift in either direction. |

---

PR #271, opened 2026-05-31

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `2ace33bc-14e0-45d9-ab49-2187a385f694`

### DIFF
--- a/app/backend/rag/catalog.py
+++ b/app/backend/rag/catalog.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import logging
 
-from backend.config import CATALOG_CACHE_TTL_SECONDS
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
@@ -78,7 +77,7 @@ def build_catalog_block(videos: list[dict], tier: str) -> dict:
 
     cache_control: dict = {"type": "ephemeral"}
     if tier == "extended":
-        cache_control["ttl"] = CATALOG_CACHE_TTL_SECONDS  # integer seconds per Anthropic API
+        cache_control["ttl"] = "1h"  # Anthropic accepts only "5m" or "1h" duration strings
 
     return {
         "type": "text",
--- a/app/backend/tests/test_catalog.py
+++ b/app/backend/tests/test_catalog.py
@@ -71,7 +71,7 @@ def test_build_catalog_block_standard() -> None:
 def test_build_catalog_block_extended() -> None:
     videos = [{"id": "vid-abc", "title": "My Video", "url": "https://youtube.com/watch?v=abc"}]
     block = catalog.build_catalog_block(videos, "extended")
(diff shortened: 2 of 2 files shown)


coleam00:

This is an excellent, tightly-scoped fix that nails the issue's core reasoning test. The implementation correctly identifies that Anthropic's cache_control.ttl accepts only the duration enums "5m"/"1h" and avoids the seductive wrong fix of stringifying the integer ("3600"). The change is surgical (two files, dead import removed, comment corrected to state the real contract) and the test update is principled: it rewrites the test that previously encoded the wrong belief rather than gaming assertions to pass. Plan and implementation are in lockstep.

The only soft spot is that CATALOG_CACHE_TTL_SECONDS now sits orphaned in config.py with no consumers — the issue itself flagged that the "TTL in seconds" abstraction is conceptually wrong and is really a tier selector. Cleaning that up (or converting it to a tier→duration mapping) would have been a slightly more complete root-cause fix, but leaving it is a reasonable scope-discipline decision and arguably belongs in a separate change. Overall a strong, production-ready PR scoring 66/70.

## Scores

|---|---|---|
| solves_issue | 10/10 | Exactly meets every acceptance criterion: extended tier now sends the valid Anthropic duration string "1h", standard tier still emits plain {"type": "ephemeral"} with no ttl, and the catalog block text/inclusion is untouched. |
| root_cause | 9/10 | Fixes the real root cause — the wrong belief (encoded in the old comment) that the API takes integer seconds — by mapping the tier to Anthropic's enum value. Minor nit: CATALOG_CACHE_TTL_SECONDS is now an orphaned config constant, which the issue hinted is itself the wrong abstraction, but leaving it untouched is a defensible scope call. |
| scope_discipline | 10/10 | Surgical: two files, +8/-10 lines, touching only the ttl line, its dead import, and the two affected tests. No unrelated churn. |
| subtle_correctness | 9/10 | Avoids the tempting wrong fix (stringifying to "3600") and uses the correct duration enum. Standard-tier path and type stay correct; no foot-guns introduced. |
| code_quality | 9/10 | Clean and idiomatic; the replacement comment now accurately states the API contract. Hardcoding "1h" rather than a tier→duration map is fine given only two tiers exist. |
| test_discipline | 9/10 | Correctly rewrites the misleading test (which asserted an integer ttl) into one asserting the Anthropic duration string, renamed appropriately, plus updates the extended-block assertion. Tests were fixed to reflect correct behavior, not gamed to pass. |
| plan_impl_fidelity | 10/10 | Implementation faithfully executes the plan in full: ttl set to "1h", unused import removed, comment corrected, and both named tests updated exactly as specified. No drift in either direction. |

---

PR #272, opened 2026-05-31

### LINKED ISSUE #247: fix(web): dynamous citation with empty lesson_url is a silent dead click

```js
    if (citation.lesson_url) {
    }
  } else {
  }
}, []);
```

### Acceptance

### Out of scope

### Why this tests reasoning

coleam00:
Fixes #247

**Source run-id:** `7333cf17-e42d-4194-800f-95409875b1a9`

### DIFF
--- a/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
+++ b/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
@@ -1,9 +1,9 @@
 /**
  * Unit tests for ChatArea.handleCitationClick branching logic (issue #216 fix).
  *
- * Dynamous citations → window.open(lesson_url, '_blank', 'noopener,noreferrer')
- * YouTube citations  → opens CitationModal (setSelectedCitation)
- * Missing lesson_url → does nothing (no blank tab)
+ * Dynamous citations       → window.open(lesson_url, '_blank', 'noopener,noreferrer')
+ * YouTube citations        → opens CitationModal (setSelectedCitation)
+ * Missing lesson_url       → informational toast (no blank tab, no modal)
  */
 
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


coleam00:

A clean, near-textbook fix. The implementation correctly identifies that the silent dead click stems from a missing else branch in the Dynamous citation routing and adds an informational toast using the already-available addToast hook, while preserving the populated-lesson_url new-tab path and the YouTube modal path. Crucially it sidesteps the reasoning trap called out in the issue — it does not fall through to the YouTube-only modal, which would render 'video unavailable' for a Dynamous chunk. Scope is tight (one handler branch plus its test), subtle correctness is handled (dependency array updated, falsy guard reused for both '' and undefined), and the test is genuinely strengthened rather than loosened to pass.

Weaknesses are minor and largely a matter of degree: the toast is a good-enough UX choice but not the richest possible fallback (e.g., rendering the chip as visibly non-interactive), and the fix is confined to the handler so a user could still hover an apparently-clickable chip before discovering it only yields a toast — though the issue explicitly accepted a toast as valid. Plan-implementation fidelity is essentially perfect, with the diff matching the planned message text, dependency change, and test structure line for line.

## Scores

|---|---|---|
| solves_issue | 10/10 | Hits all three acceptance criteria exactly: empty lesson_url now fires an informational toast (feedback instead of dead click), populated lesson_url still opens a new tab, and YouTube citations still route to the modal. It also avoids the tempting wrong fix of falling through to setSelectedCitation. |
| root_cause | 9/10 | Fixes the actual gap — the missing else branch in the source_type routing — rather than papering over it, and correctly respects the YouTube-only contract of CitationModal. Not the deepest possible fix (a non-interactive chip), but the issue explicitly authorized a toast as a sensible fallback. |
| scope_discipline | 10/10 | Surgical: only handleCitationClick (one else branch + dependency array), its comment block, and the corresponding test file. No unrelated churn, no modal or backend changes. |
| subtle_correctness | 9/10 | Correctly adds addToast to the useCallback dependency array (avoids a stale-closure/lint foot-gun), relies on the existing falsy guard that covers both "" and undefined, and reuses the established 'info' toast type. No introduced bugs. |
| code_quality | 9/10 | Idiomatic and consistent with existing addToast usage already present in the component; clear, accurate comment update. Reads like the surrounding code. |
| test_discipline | 9/10 | Test is meaningfully updated, not gamed: it refactors the useToast mock to a module-level spy, asserts the toast fires once with the right message regex and 'info' type, and adds negative assertions that no spurious toast fires on the happy/YouTube paths. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan essentially verbatim — same else-branch message text, same dependency-array change, same comment update, and the exact planned test refactor (module-level mockAddToast, regex matcher, happy-path no-toast checks). |

---

PR #273, opened 2026-05-31
[bench-OO] fix(web): give feedback for Dynamous citation with empty lesson_url

### LINKED ISSUE #247: fix(web): dynamous citation with empty lesson_url is a silent dead click

coleam00:
Fixes #247

**Source run-id:** `681c1241-3257-4696-a643-53bb2523a317`

### DIFF
--- a/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
+++ b/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
@@ -3,7 +3,7 @@
  *
  * Dynamous citations → window.open(lesson_url, '_blank', 'noopener,noreferrer')
  * YouTube citations  → opens CitationModal (setSelectedCitation)
- * Missing lesson_url → does nothing (no blank tab)
+ * Missing lesson_url → fires an info toast (no blank tab, no modal)
  */
 
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
@@ -44,8 +44,9 @@ vi.mock('../hooks/useStreamingResponse', () => ({
   }),
 }));
 
+const mockAddToast = vi.hoisted(() => vi.fn());
 vi.mock('../hooks/useToast', () => ({
-  useToast: () => ({ addToast: vi.fn(), removeToast: vi.fn() }),
+  useToast: () => ({ addToast: mockAddToast, removeToast: vi.fn() }),
 }));
 
 vi.mock('../hooks/useAuth', () => ({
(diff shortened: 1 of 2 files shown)


coleam00:

A clean, high-quality fix that resolves issue #247 exactly as specified. The change adds an informational toast in the previously-empty else branch of the Dynamous citation path, correctly avoiding the tempting (and wrong) fall-through to the YouTube-only CitationModal, which the issue explicitly flags as the reasoning trap. Scope is surgical — only the handler and its co-located test are touched — and subtle correctness is handled well: the truthiness check covering both "" and undefined is preserved, and the useCallback dependency array is updated for addToast. Test discipline is strong: the existing 'does nothing' test is meaningfully rewritten rather than gamed, with positive assertion of the info toast and negative assertions on the other two paths.

The implementation tracks the source plan virtually line-for-line, indicating tight plan/impl coupling for this OO (plan=Opus, impl=Opus) cell. Weaknesses are minor and mostly inherent to the small surface: a toast is the lightest acceptable feedback rather than a more discoverable affordance (e.g., a disabled/annotated chip), but that heavier option was correctly judged out of scope. Overall an exemplary small fix with no notable defects.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly satisfies all three acceptance criteria: empty lesson_url now fires an info toast (feedback, not a silent no-op), populated lesson_url still opens a new tab, and YouTube citations still open the modal. |
| root_cause | 9/10 | Fixes the actual gap — the empty else branch — and correctly avoids the tempting modal fall-through, choosing a toast precisely because CitationModal is YouTube-only. A genuine structural fix, not a band-aid. |
| scope_discipline | 10/10 | Touches only the handler and its test (17 add / 5 del across 2 files). No modal, backend, or chip-rendering changes — exactly the surface the issue scoped. |
| subtle_correctness | 9/10 | Preserves the `if (citation.lesson_url)` truthiness check so both "" and undefined route to the toast, and correctly adds `addToast` to the useCallback dependency array. No foot-guns introduced. |
| code_quality | 9/10 | Reuses the already-in-scope addToast and existing 'info' toast type, with a clear comment explaining why the modal fall-through is wrong. Idiomatic and readable; message wording is sensible. |
| test_discipline | 9/10 | Hoists the addToast mock to make it assertable, renames the stale test, asserts the toast fires with 'info', and adds negative assertions that no toast fires on the working-link and modal paths. Targeted and meaningful, not gamed. |
| plan_impl_fidelity | 10/10 | Implementation is essentially identical to the plan — same else branch, same message string, same dependency-array change, same test hoist and assertions. No drift. |

---

PR #274, opened 2026-05-31

### LINKED ISSUE #247: fix(web): dynamous citation with empty lesson_url is a silent dead click

coleam00:
Fixes #247

**Source run-id:** `2f696be6-44c1-4f93-a394-c06da6cd6aac`

### DIFF
--- a/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
+++ b/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
@@ -3,7 +3,7 @@
  *
  * Dynamous citations → window.open(lesson_url, '_blank', 'noopener,noreferrer')
  * YouTube citations  → opens CitationModal (setSelectedCitation)
- * Missing lesson_url → does nothing (no blank tab)
+ * Missing lesson_url → info toast (no blank tab)
  */
 
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
@@ -44,8 +44,10 @@ vi.mock('../hooks/useStreamingResponse', () => ({
   }),
 }));
 
+const addToastRef = { current: vi.fn() };
+
 vi.mock('../hooks/useToast', () => ({
-  useToast: () => ({ addToast: vi.fn(), removeToast: vi.fn() }),
+  useToast: () => ({ addToast: addToastRef.current, removeToast: vi.fn() }),
 }));
 
 vi.mock('../hooks/useAuth', () => ({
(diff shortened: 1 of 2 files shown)


coleam00:

A clean, correct, and tightly-scoped fix that lands exactly on the issue's intended solution. handleCitationClick gains an else-branch that fires an 'info' toast when a Dynamous citation has an empty lesson_url, while preserving new-tab behavior for populated links and the modal path for YouTube citations. It correctly sidesteps the trap the issue called out (falling through to the YouTube-only modal) and shows good React hygiene by adding addToast to the dependency array. Test discipline is strong: the previously-passing 'does nothing' assertion is honestly rewritten to the new contract, and a separate empty-string case is added, both asserting the exact toast call and the absence of a blank tab or modal.

Plan-implementation fidelity is high — the diff matches the plan's approach, file targets, message copy, and test scenarios almost exactly, with only trivial drift in the test file name and an unplanned (but correct) useCallback dependency fix. There are no real weaknesses; this is a production-grade change for the KK (plan=Kimi, impl=Kimi) cell.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly satisfies all three acceptance criteria: an empty lesson_url now surfaces an info toast instead of a silent no-op, a populated lesson_url still opens in a new tab, and YouTube citations still route to the modal via setSelectedCitation. |
| root_cause | 9/10 | Fixes the real gap in the source_type routing contract by handling the dynamous-with-no-link case in the click handler — the sanctioned toast fallback the issue recommended. It deliberately avoids the tempting wrong fix of falling through to the YouTube-only modal, which would render 'video unavailable'. |
| scope_discipline | 10/10 | Touches only ChatArea.tsx (the handler) and its test file. No unrelated edits, no churn elsewhere — exactly the surgical scope the issue's out-of-scope list demanded. |
| subtle_correctness | 9/10 | Correctly relies on the existing truthiness guard to catch both "" and undefined, properly adds addToast to the useCallback dependency array to avoid a stale closure, and uses the 'info' toast type which the exploration confirms ToastProvider supports. No async or type foot-guns. |
| code_quality | 9/10 | Clean, idiomatic, mirrors the surrounding pattern; clear user-facing message and a minimal else-branch. Reads like the existing code and follows the codebase's established toast-feedback convention. |
| test_discipline | 9/10 | Legitimately rewrites the stale 'does nothing' test to the new contract and adds a dedicated test for the empty-string case. Captures addToast via a ref mock reset in beforeEach and asserts exact call args, no window.open, and no modal — meaningful and not gamed. |
| plan_impl_fidelity | 9/10 | The implementation faithfully executes the plan: the single else-branch info toast in handleCitationClick with the exact suggested copy, plus tests for the empty/missing lesson_url paths. Only minor drift is the test file name (the plan named ChatArea.test.tsx; the change lives in the existing ChatArea-handleCitationClick.test.tsx), and the useCallback dep-array fix was an unplanned-but-correct detail. |

---

PR #275, opened 2026-05-31
[bench-KO] review: address self-review findings

### LINKED ISSUE #247: fix(web): dynamous citation with empty lesson_url is a silent dead click

coleam00:
Fixes #247

**Source run-id:** `8fba76cd-aa55-4c18-a502-82f7cf296c2a`

### DIFF
--- a/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
+++ b/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
@@ -172,7 +172,7 @@ describe('handleCitationClick', () => {
     expect(screen.queryByTitle('YouTube video player')).not.toBeInTheDocument();
     // ...instead the user gets an info toast rather than a silent dead click
     expect(mockAddToast).toHaveBeenCalledTimes(1);
-    expect(mockAddToast).toHaveBeenCalledWith({ message: expect.any(String), type: 'info' });
+    expect(mockAddToast).toHaveBeenCalledWith(expect.any(String), 'info');
   });
 
   it('opens the citation modal for YouTube citations', async () => {
--- a/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
+++ b/app/frontend/src/__tests__/ChatArea-handleCitationClick.test.tsx
@@ -15,6 +15,9 @@ import type { Citation, Message } from '../lib/api';
 // ── Shared mocks ──────────────────────────────────────────────────────────────
 
 const mockNavigate = vi.fn();
+// Captured at module scope (mock-prefixed so vi.mock hoisting allows the
+// reference) so tests can assert on the toast surfaced by handleCitationClick.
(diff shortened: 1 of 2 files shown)


coleam00:

A clean, surgical fix that fully solves the issue and demonstrates genuine reasoning about the source_type routing contract. It correctly identifies that the citation modal is YouTube-only and therefore cannot be used as a fallback for a Dynamous source with no link, choosing an info toast instead — exactly the kind of cross-cutting understanding the issue was designed to test. Scope is tight (two files, small diff), subtle correctness is strong (right addToast signature, correct dependency array), and the test is upgraded meaningfully rather than gamed.

The standout signal is plan/impl fidelity in the right direction: the Kimi-produced plan called addToast with an object argument that would not have compiled against the (message, type) positional signature; the Opus implementation silently corrected this to the proper positional call and matched the test assertion accordingly. Weaknesses are minor and cosmetic — the toast copy is an inline literal rather than a named constant — and do not detract from a production-grade result. Strong cell overall.

## Scores

|---|---|---|
| solves_issue | 10/10 | Hits all three acceptance criteria: empty lesson_url now surfaces an info toast (no longer a silent dead click), populated lesson_url still opens a new tab, and YouTube citations still route to the modal. It also avoids the tempting wrong fix of falling through to setSelectedCitation. |
| root_cause | 9/10 | Correctly recognizes the real contract — the citation modal is YouTube-only and would render 'video unavailable' for a Dynamous chunk — and chooses a sensible non-modal fallback rather than papering over the symptom. The inline comment documents the source_type routing assumption it depends on. |
| scope_discipline | 10/10 | Touches only the two files the issue implicates (ChatArea.tsx handler + its test). +25/-10 lines, no unrelated edits, no modal/backend changes — exactly the out-of-scope boundaries respected. |
| subtle_correctness | 10/10 | Uses the actual addToast(message, type) positional signature (verified against useToast.ts), correctly adds addToast to the useCallback dependency array, and preserves the window.open noopener flags. No foot-guns introduced. |
| code_quality | 9/10 | Readable, idiomatic to the codebase, and the comment explains the why (YouTube-only modal) rather than the what. Minor: the toast string is inline rather than a constant, but that matches existing conventions. |
| test_discipline | 9/10 | The existing 'does nothing' test is meaningfully upgraded, not gamed: it still asserts window.open is not called and the modal does not open, and additionally asserts addToast was called once with (string, 'info'). Module-scoped mock is wired cleanly for the vi.mock hoist. |
| plan_impl_fidelity | 9/10 | Implementation faithfully executes the plan's intent (toast in empty branch, dep-array update, test assertion change). The one divergence is a correction: the plan specified an object-form addToast({message,type}) call that does not match the real (message, type) signature; the impl used the correct positional form. |

---

Issue #208, opened 2026-04-27
fix(citations): collapse 40+ same-video chunks into a single citation per video


coleam00:
## Symptom

When a single video answers a question, the response shows **~40 citation chips** that are all 5–10 second clips from the same video. The user sees a wall of chips like:

> Video Foo (0:00) | Video Foo (0:08) | Video Foo (0:14) | Video Foo (0:23) | … (×40)

Instead of a single chip pointing to one timestamp (or a chip with a count, like *Video Foo (3 segments)*).

This happens because a single video is chunked into many ~10-second pieces during ingest, the model legitimately cites several of them to back consecutive sentences, and the frontend renders each cited chunk as a standalone chip with no per-video grouping.

## Expected

When multiple chunks from the same video are cited, render **one chip per video**, ideally pointing to the earliest cited timestamp (or the most-relevant one), with some visual indication that more segments back the answer.

Across-video diversity should be preserved — e.g. if 3 videos are cited, show 3 chips, not the cross-product.

## File pointers (commit \`476a17f\`)

- \`app/backend/llm/openrouter.py:71-107\` — system prompt citation rules. They tell the model to emit \`[c:<chunk_id>]\` for *every* grounded sentence and to *stack* same-sentence multi-chunk markers. There's no instruction about per-video density.
- \`app/backend/routes/messages.py:198-217\` — dedup is by \`chunk_id\` only; \`is_cited\` is set; cap is \`CITATIONS_MAX_COUNT\`. No video-level grouping.
- \`app/backend/rag/expansion.py:101-145\` — \`expand_and_merge\` *does* merge consecutive chunks into spans, but only during retrieval, not in the citation path.

## Possible fix directions

There are three reasonable approaches; pick one or stack them:

### Option 1: Group at render time (frontend, lowest risk)
In the chat \`Message\` component, group citations by \`video_id\`, render one chip per group, label it with the video title, and on click open the earliest cited timestamp. Optionally show *N segments* below the title. **Pros:** no model or backend changes. **Cons:** the LLM still emits 40 markers — wasted tokens.

### Option 2: Collapse server-side before sending to the client
[...]

PR #282, opened 2026-05-31

### LINKED ISSUE #276: Missing source citation when an answer uses two nearby moments from the same video

coleam00:
Fixes #276

**Source run-id:** `73797838-98ad-4578-b8d9-cbdc14a5d980`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -211,9 +211,9 @@ async def create_message(
                         cited_ids = extract_cited_chunk_ids(final_text_raw)
                         for chunk in source_citations:
                             chunk["is_cited"] = chunk.get("chunk_id") in cited_ids
-                        # Collapse same-video chunks (issue #208): keep one
-                        # entry per video_id, choosing the earliest-cited
-                        # timestamp as the representative.
+                        # Collapse same-video chunks: each cited chunk gets its
+                        # own chronological chip (issue #276); uncited chunks for
+                        # a video still collapse to one consulted chip (#208).
                         source_citations[:] = _collapse_by_video(source_citations)
                         # Cap fallback (issue #176): cited pass through, non-cited sliced.
                         cited = [c for c in source_citations if c.get("is_cited")]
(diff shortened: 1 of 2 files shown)


coleam00:

This is an excellent, tightly-scoped fix. It correctly identifies that `_collapse_by_video()` was the root cause of both symptoms — collapsing every cited chunk in a video to a single earliest-timestamp representative — and replaces that branch with a fan-out that emits one chronologically-ordered chip per cited chunk while preserving the issue #208 uncited-collapse behavior. The choice to key the fan-out on whether a chunk was cited (rather than on timestamp distance) is the structurally right call and avoids reintroducing the bug. The cap, dedup, refusal, and persistence paths are correctly left untouched, and the implementation is a near-verbatim execution of a strong plan.

Weaknesses are minor. Uncited chunks belonging to a video that also has cited chunks are dropped from the consulted tier entirely (intentional per the plan, but a subtle behavioral edge worth noting). The tests are meaningful and target the exact regression rather than being gamed. Overall a high-quality, production-grade PR with strong plan/impl fidelity.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes both reported symptoms: every distinct cited moment from the same video now gets its own chip, and each keeps its own correct timestamp so the later moment is no longer dropped in favor of the earlier one. Covers both live-streaming and reload paths since collapse happens before persistence. |
| root_cause | 9/10 | Targets the actual cause — `_collapse_by_video()` discarding all but the earliest representative for cited groups — rather than a heuristic. Correctly keys the fan-out on whether a chunk was cited (not on timestamp proximity), avoiding a distance-threshold band-aid that would reintroduce the drop. |
| scope_discipline | 10/10 | Surgical: only the cited-group branch of `_collapse_by_video()` plus its docstring/comment, and the test file. No frontend, DB, or unrelated changes — exactly the two files the issue requires. |
| subtle_correctness | 8/10 | Sorts cited chunks chronologically (not relying on insertion order), sets segment_count=1 per cited chip, leaves the #176 cap logic intact (cited entries pass through uncapped), and preserves the uncited #208 collapse. Minor: uncited chunks within a cited video are absorbed entirely (no consulted chip for them), which is intentional per the plan but is a slight behavioral nuance. |
| code_quality | 9/10 | Clean, idiomatic Python; the rewritten docstring and inline comment clearly document the two-tier contract and cite the relevant issues (#276/#208). Reads consistently with the surrounding citation-building code. |
| test_discipline | 9/10 | Adds a precise regression test asserting two distinct cited moments survive with both timestamps present, chronological ordering, and segment_count==1 — directly covering the missing-chip and wrong-timestamp failure modes. Existing tests were updated to the new (correct) semantics rather than gamed; the uncited-collapse test is left untouched. |
| plan_impl_fidelity | 10/10 | The implementation is essentially a verbatim execution of source-plan.md: same branch rewrite (one entry per sorted cited chunk, segment_count=1), same docstring update, same primary regression test plus the three planned test updates, and the planned out-of-scope items were all respected. |

---

PR #283, opened 2026-05-31
[bench-OK] review: fix stale call-site comment for _collapse_by_video

### LINKED ISSUE #276: Missing source citation when an answer uses two nearby moments from the same video

coleam00:
Fixes #276

**Source run-id:** `657e778d-cc5c-42a2-8eaf-cf9fa1b61a29`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -211,9 +211,9 @@ async def create_message(
                         cited_ids = extract_cited_chunk_ids(final_text_raw)
                         for chunk in source_citations:
                             chunk["is_cited"] = chunk.get("chunk_id") in cited_ids
-                        # Collapse same-video chunks (issue #208): keep one
-                        # entry per video_id, choosing the earliest-cited
-                        # timestamp as the representative.
+                        # Collapse same-video chunks (issue #208/#276): videos
+                        # with ≥2 cited moments fan out to one chip each;
+                        # single-cited and uncited groups collapse to one entry.
                         source_citations[:] = _collapse_by_video(source_citations)
                         # Cap fallback (issue #176): cited pass through, non-cited sliced.
                         cited = [c for c in source_citations if c.get("is_cited")]
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -467,22 +467,22 @@ async def _maybe_set_conversation_title(
 
 
 def _collapse_by_video(chunks: list[dict[str, Any]]) -> list[dict[str, Any]]:
-    """Collapse multiple chunks from the same video into a single citation entry.
+    """Collapse multiple chunks from the same video into citation entries.
 
     After the ``is_cited`` pass, chunks from the same video are redundant in
     the UI — the user needs one clickable chip per video, not one per 5-second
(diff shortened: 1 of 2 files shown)


coleam00:

An excellent, tightly-scoped fix. It correctly diagnoses the root cause — _collapse_by_video collapsing every same-video chunk into a single earliest-cited representative — and replaces it with a clean three-branch policy that fans out videos with two or more cited moments into one correctly-timestamped chip each, while preserving the issue #208 anti-flooding behaviour for single-cited and uncited groups. Subtle details are handled thoughtfully: dedup by chunk_id rather than timestamp, timeline ordering within a group, None-safe sorting, segment_count semantics, and untouched issue #176 cap logic. Tests are targeted and exercise the real SSE path, including the identical-timestamp edge case. Implementation fidelity to the Opus plan is near-perfect, and the Kimi self-review pass appropriately tidied the now-stale call-site comment (the PR title reflects only that review commit, which understates the substantive fix in the first commit). Weaknesses are negligible — only minor branch boilerplate duplication and the cosmetic title/scope-labelling mismatch from the benchmark harness.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes both reported symptoms: two nearby cited moments from the same video now each emit their own chip, and the surviving chip no longer collapses to the wrong (earliest) timestamp. Persistence round-trips the multi-entry list, so it holds both live and on reload as the issue required. |
| root_cause | 9/10 | Targets the actual structural cause — _collapse_by_video unconditionally reducing every same-video group to one earliest-cited representative — rather than masking it. The intermittency ('nearby moments only') is fully explained by this collapse, and the fan-out branch removes it. |
| scope_discipline | 10/10 | Surgical: one function rewrite plus targeted tests across exactly 2 files, +63/-21. No frontend, schema, or unrelated changes; the only extra touch (the stale call-site comment) is squarely in-scope and was the self-review commit's stated purpose. |
| subtle_correctness | 9/10 | Handles the tricky cases well: dedup by chunk_id (not timestamp) so identical-timestamp moments still split, intra-group ordering by start_seconds, None-safe `or 0.0`, segment_count=1 to suppress the '(N segments)' suffix on fan-out chips, and the issue #176 cap logic is preserved since cited entries pass through uncapped. |
| code_quality | 9/10 | Readable, idiomatic to the module, clear three-branch structure, and an accurate rewritten docstring documenting the new contract. Minor nit: the three branches each repeat the dict-copy/append boilerplate, but it stays clear over clever. |
| test_discipline | 9/10 | Two meaningful regression tests through the real SSE path reusing existing helpers: distinct-timestamp fan-out (the core regression) and the same-timestamp/dedup-by-chunk_id edge case. Asserts ids, timestamps, is_cited, and segment_count. Existing collapse tests remain unmodified. |
| plan_impl_fidelity | 10/10 | The diff executes the plan almost verbatim: the exact three branches (≥2 cited fan-out with segment_count=1 sorted by start_seconds, 1-cited representative, 0-cited earliest), the docstring update, both named tests with the planned fixtures, no frontend/call-site logic change, and the explicitly-deferred comment cleanup done in the self-review pass. |

---

PR #284, opened 2026-05-31
[bench-KO] review: address self-review findings

### LINKED ISSUE #276: Missing source citation when an answer uses two nearby moments from the same video

coleam00:
Fixes #276

**Source run-id:** `615c0fc6-5d3f-4d06-b91c-8b7b84fdbe46`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -466,10 +466,7 @@ async def _maybe_set_conversation_title(
         await repository.update_conversation_title(conv_id, user_id=user_id, title=title)
 
 
-# Cited chunks from the same video are merged into one chip only when their
-# start timestamps fall within this many seconds of each other. Moments farther
-# apart in the same video get their own chip (issue #276) so every distinct
-# moment the assistant actually used is shown with its own correct timestamp.
+# Cited moments farther apart than this (seconds) get separate chips (issue #276).
 CITED_CLUSTER_GAP_SECONDS = 60.0
 
 
@@ -505,8 +502,6 @@ def _collapse_by_video(chunks: list[dict[str, Any]]) -> list[dict[str, Any]]:
 
     cited_entries: list[dict] = []
     for group in _group_by_video(cited).values():
-        # Cluster by start_seconds proximity so nearby moments merge but
-        # distant moments each get their own chip.
         ordered = sorted(group, key=lambda c: c.get("start_seconds") or 0.0)
         cluster: list[dict] = []
         prev_start: float | None = None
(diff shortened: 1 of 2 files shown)


coleam00:

A strong, disciplined fix. It correctly locates the root cause in _collapse_by_video — the unconditional per-video collapse that flattened distinct cited moments into a single earliest-timestamp chip — and repairs it at exactly the layer that feeds both the live SSE stream and the persisted message, so live and reload behaviour are fixed together. Scope is tight (the two planned files only), the refactor into small helpers is clean and well-documented with issue references, and the test suite meaningfully covers both the #276 regression and the #208 anti-clutter guard. Plan-to-implementation fidelity is essentially perfect, with the implementer adding a sensible ordering test and helper extraction on top.

The main weakness is the 60-second proximity heuristic. It is a reasonable proxy for distinguishing 'multiple segments of one moment' from 'two distinct moments,' but it is a magic threshold rather than a principled signal, and it means the literal 'nearby' case from the issue title is only fixed when the two moments happen to be >60s apart; tighter pairs still merge. The single-linkage clustering (gap measured against the previous chunk, not the cluster origin) is a minor latent foot-gun. Neither is disqualifying — both are defensible engineering trade-offs given the competing #208 requirement — but they keep solves_issue and subtle_correctness just short of top marks.

## Scores

|---|---|---|
| solves_issue | 8/10 | Directly addresses both reported symptoms: distinct same-video moments now each get a chip (cited chunks no longer collapse to one entry), and each chip carries its own earliest-in-cluster timestamp, fixing the wrong-timestamp case. Because the collapse runs before both the SSE emit and the DB write, it fixes live streaming and reload parity at once. Caveat: the fix relies on a 60s proximity threshold, so two genuinely distinct moments closer than 60s apart would still merge — and the issue title literally says 'nearby' — leaving the very tightest case theoretically unfixed. |
| root_cause | 9/10 | Correctly identifies that the unconditional per-video collapse was destroying distinct cited moments and always picking the earliest representative — the structural cause of both the missing chip and the wrong timestamp. The fix repairs the collapse logic itself at the right layer rather than patching the frontend or SSE format. The 60s gap is a heuristic proxy for 'same moment vs. distinct moment' rather than a first-principles signal, which is the only thing keeping this from a 10. |
| scope_discipline | 9/10 | Surgical: only messages.py (the target helper) and its test file, exactly the two files the plan named, with a clean refactor into _group_by_video/_make_entry helpers and no unrelated churn. Worth flagging that messages.py is on the protected-paths list, but the change touches the citation-display helper, not the auth/rate-limit invariants, so it's process-adjacent rather than a real scope violation. |
| subtle_correctness | 8/10 | Handles None start_seconds via 'or 0.0', preserves first-seen order, sorts before clustering, and keeps the cited-before-uncited convention. One subtle foot-gun: clustering compares each chunk to the previous chunk's start (single-linkage), so a chain of chunks each <60s apart merges into one cluster spanning well over 60s total — though top-k=5 retrieval bounds this in practice. No async/type regressions. |
| code_quality | 9/10 | Readable, idiomatic, well-named helpers and a clearly-documented module constant; docstrings explicitly tie behaviour to issues #208 and #276. Type hints present throughout, though a couple of helper annotations use bare list[dict] rather than the file's list[dict[str, Any]] convention — a trivial nit. |
| test_discipline | 9/10 | Four targeted, meaningful tests: the #276 regression (two distinct moments → two chips with correct timestamps), the #208 guard (adjacent chunks still merge), the unchanged uncited path, and cited-before-uncited ordering. They exercise the real helper, not mocks of it. A boundary test at exactly 60s would have been the only meaningful addition. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — cited/uncited split, 60s threshold, sort-then-walk clustering, cited-first concatenation, and the three named tests — and even adds value beyond it (a fourth ordering test plus extraction of _group_by_video/_make_entry helpers consistent with the plan's intent). No drift, no dropped items. |

---

PR #285, opened 2026-05-31

### LINKED ISSUE #276: Missing source citation when an answer uses two nearby moments from the same video

coleam00:
Fixes #276

**Source run-id:** `3635c07e-bf00-4e86-ab7f-dcba2f970604`

### DIFF
--- a/app/backend/config.py
+++ b/app/backend/config.py
@@ -135,6 +135,13 @@ RETRIEVAL_MAX_PER_VIDEO: int = int(os.environ.get("RETRIEVAL_MAX_PER_VIDEO", "3"
 # Cap on non-cited citations (issue #176); cited chunks always pass through.
 CITATIONS_MAX_COUNT: int = int(os.environ.get("CITATIONS_MAX_COUNT", "10"))
 
+# Time-proximity window for collapsing same-video chunks into one citation.
+# Chunks whose start times are within this many seconds merge into a single
+# chip; farther apart remain separate (issue #276).
+CITATION_COLLAPSE_PROXIMITY_SECONDS: int = int(
+    os.environ.get("CITATION_COLLAPSE_PROXIMITY_SECONDS", "30")
+)
+
 # RAG tool-based retrieval — the LLM drives retrieval via tool calls
 # (search_videos, keyword_search_videos, semantic_search_videos,
 # get_video_transcript) rather than receiving pre-retrieved chunks. Disabled
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -28,7 +28,12 @@ from pydantic import BaseModel, Field, field_validator
 
 from backend import rate_limit
 from backend.auth.dependencies import get_current_user
-from backend.config import CITATIONS_MAX_COUNT, LLM_TOOLS_ENABLED, LLM_TOOLS_MAX_PER_TURN
+from backend.config import (
+    CITATION_COLLAPSE_PROXIMITY_SECONDS,
+    CITATIONS_MAX_COUNT,
+    LLM_TOOLS_ENABLED,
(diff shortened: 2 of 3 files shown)


coleam00:
**Benchmark Evaluator** — score: **51/70**

A competent, well-scoped fix that reframes the same-video citation collapse (issue #208) as proximity-aware clustering: chunks within a configurable 30s window merge into one chip while farther-apart moments each keep a correctly-timestamped chip. The code is clean, idiomatic, env-configurable, and the interval-merge logic (running max end, sorted starts, None-safe) is correct. Scope is tight — config constant, the collapse function, and tests.

The core weakness is that the fix only partially solves the reported problem. The issue is explicitly about 'nearby'/'close together' moments, yet anything inside the 30s window is still merged AND still resolves to the earliest-cited representative, so both reported symptoms (missing chip and wrong/earlier timestamp) can survive for sub-threshold gaps; the threshold is an arbitrary heuristic rather than a true distinct-moment detector. Testing reinforces this: the new test actually exercises the far-apart case (and is misleadingly named 'nearby'), with no coverage of the truly-nearby or wrong-timestamp scenarios. Plan fidelity is unscorable as the planning artifact was absent.

## Scores

|---|---|---|
| solves_issue | 6/10 | Introduces a proximity-based clustering so distinct same-video moments more than ~30s apart now each get their own correctly-timestamped chip, which addresses the general class of bug. But the issue specifically describes 'nearby' moments 'close together' — and within the 30s window the fix still merges them into one chip and still picks the earliest-cited representative, so the literal reported scenario (and its wrong-timestamp symptom) can remain unfixed for sub-threshold gaps. |
| root_cause | 7/10 | Correctly identifies that issue #208's unconditional same-video collapse is the structural cause of distinct moments losing their chips, and fixes that function directly rather than patching the symptom elsewhere. Loses points because the remedy is an arbitrary 30s heuristic that approximates 'adjacent segment vs distinct moment' rather than truly detecting distinct moments. |
| scope_discipline | 8/10 | Surgical and on-target: one new config constant, the rewrite of _collapse_by_video, and the corresponding tests — three relevant files, no unrelated churn. Minor note: messages.py is on the protected-paths list, though the modified logic is citation collapse, not the auth/rate-limit invariants that motivate the protection. |
| subtle_correctness | 7/10 | The greedy interval merge is sound — it sorts by start, tracks the running max end (cluster_ends[-1]) so overlapping/nested chunks merge correctly, and handles None start/end via 0.0 fallback. However, the within-cluster representative still picks the earliest cited chunk, so the 'wrong timestamp points at the earlier moment' symptom persists whenever two cited moments fall inside the same cluster. |
| code_quality | 8/10 | Clean, readable, idiomatic to the codebase: env-configurable constant with a clear comment, accurately updated docstring, explicit group_order tracking, and a straightforward clustering loop. Could have been factored into a helper, but it reads well as-is. |
| test_discipline | 6/10 | Adds a focused unit test asserting two separated chunks yield two distinct correctly-timestamped chips, and sensibly retunes the #208 collapse test (5s spacing) to stay within one cluster. Weaknesses: the new test is named 'nearby' but actually exercises the FAR-apart (gap = threshold+10) case, and nothing covers two genuinely-nearby cited moments or the wrong-timestamp symptom the issue called out. |
| plan_impl_fidelity | 9/10 | The implementation is a near-verbatim execution of source-plan.md: the CITATION_COLLAPSE_PROXIMITY_SECONDS=30 constant, the greedy per-video clustering with the exact 'start <= cluster_end + threshold' / 'cluster_end = max(...)' merge rule, the retuned #208 test, and the new test named exactly as planned. The plan's stated out-of-scope items (no frontend/SSE/retriever changes) were respected. Slightly short of 10 because the exploration flagged the residual 'wrong representative timestamp within a cluster' risk, which neither the plan nor the impl actually closes. |

---

Issue #277, opened 2026-05-31
Saved answer's sources don't match what was shown live after an interrupted response


coleam00:
When the assistant has trouble part-way through an answer (this tends to happen on longer answers or when the upstream model is flaky), I sometimes end up with a saved assistant message whose sources are inconsistent with what I saw while it was streaming.

On reload, the saved message shows a different and longer set of source chips than what appeared live, including duplicate chips for the same video, and sometimes source chips attached to an answer that actually declined to answer. The live view and the reloaded view disagree.

Normal answers that complete successfully are consistent between the live view and reload. The problem only shows up when the answer was interrupted or errored mid-stream.

Expected: whatever sources are persisted with an answer should match what the user saw, deduplicated and finalized the same way, regardless of whether the answer completed cleanly or was interrupted.

Please fix the root cause and add a regression test.

PR #286, opened 2026-05-31

coleam00:
Fixes #277

**Source run-id:** `f2080579-0549-4e5f-8337-4696b885061a`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -17,6 +17,7 @@ Orchestrates the tool-driven RAG flow:
 from __future__ import annotations
 
 import asyncio
+import copy
 import json
 import logging
 from collections.abc import AsyncGenerator
@@ -171,6 +172,7 @@ async def create_message(
     async def event_generator() -> AsyncGenerator[str, None]:
         full_response: list[str] = []
         final_text_buf: list[str] = []
+        emitted_sources: list[dict] | None = None
         # Two-tier citations (issue #176): strip `[c:<id>]` markers from the
         # stream; use them at [DONE] to flag is_cited on retrieved chunks.
         marker_stripper = CitationMarkerStripper()
@@ -227,6 +229,7 @@ async def create_message(
                             else _extract_text_from_sse(full_response)
                         )
                         if not _is_refusal(final_text):
+                            emitted_sources = copy.deepcopy(source_citations)
                             sources_json = json.dumps(source_citations)
                             yield f"event: sources\ndata: {sources_json}\n\n"
                     full_response.append(sse_chunk)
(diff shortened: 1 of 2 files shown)


coleam00:

A strong, focused PR. It correctly diagnoses the root cause — the finally block persisting an in-place-mutated source_citations list that could be half-processed and inconsistent with the (never-emitted) live SSE event after an interrupted/errored [DONE] branch — and fixes it by snapshotting the exact emitted payload and falling back to None when no sources were shown. The change is surgical (two files, a handful of lines), idiomatic, and well-aligned with the existing refusal-suppression logic, so live and reloaded views now agree across clean, interrupted, refusal, and no-source cases.

Weaknesses are minor: the finally block keeps a now-redundant _is_refusal check with a slightly divergent text fallback (harmless but a small readability/consistency wrinkle), and one test asserts copy.deepcopy was called, coupling it to the implementation rather than pure behavior. Plan-to-implementation fidelity is essentially perfect. Overall this is high-quality work that solves the stated problem at the root with a proper regression test.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly addresses the reported inconsistency: by persisting only the snapshot of sources that was actually emitted live (and falling back to None when the stream is interrupted before emission), the reloaded view now matches what the user saw. Covers the exact failure mode in the issue (longer/duplicate chip set on reload after an interrupt). |
| root_cause | 9/10 | Correctly identifies the structural cause — the mutable source_citations list is mutated in-place during [DONE] processing (dedup, is_cited, _collapse_by_video, capping) and the finally block persisted that same partially-processed reference even when the sources SSE was never emitted. Snapshotting at the emission point is a genuine root-cause fix, not a symptom patch. |
| scope_discipline | 9/10 | Surgical: one import, one snapshot variable, one assignment, and a two-line change in the finally condition, plus added tests. No unrelated edits. messages.py is a protected path, but touching it is unavoidable for this server-side bug and the change stays well clear of the auth/rate-limit logic. |
| subtle_correctness | 8/10 | Snapshot is taken inside the non-refusal guard right before json.dumps, so emitted_sources is non-None only when sources were actually shown — keeping live and persisted paths aligned across interrupt, refusal, and no-tool cases. deepcopy is defensively correct. Minor: the finally retains a now-redundant _is_refusal check whose text-fallback differs slightly from the live check, though harmless. |
| code_quality | 8/10 | Clean, idiomatic, well-commented in the surrounding style; the snapshot variable name (emitted_sources) is clear and the typing (list[dict] | None) matches conventions. The leftover redundant refusal check in finally is a small readability nit but is defensible as defensive coding. |
| test_discipline | 9/10 | Two meaningful targeted tests added with no modification to existing tests: one reproduces the exact interrupt (raising _collapse_by_video) and asserts sources=None is persisted, the other asserts the persisted sources deep-equal the SSE payload. The deepcopy-was-called assertion is slightly implementation-coupled, but the interrupt regression test is precisely on point. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — the import copy, emitted_sources=None initializer, deepcopy snapshot immediately before the sources SSE yield, the finally swap to emitted_sources, and both named tests (interrupted-persists-None and snapshot-matches-persisted) all appear exactly as specified. |

---

PR #287, opened 2026-05-31

coleam00:
Fixes #277

**Source run-id:** `f1726af3-396a-4ffd-9820-fb3bb74ee24a`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -171,6 +171,11 @@ async def create_message(
     async def event_generator() -> AsyncGenerator[str, None]:
         full_response: list[str] = []
         final_text_buf: list[str] = []
+        # The exact sources list emitted in the live `event: sources` SSE.
+        # Stays None unless we actually emit (refusal / empty / interrupted →
+        # None). The finally block persists THIS verbatim so the saved message
+        # always matches what streamed to the client (issue #277).
+        persisted_sources: list[dict] | None = None
         # Two-tier citations (issue #176): strip `[c:<id>]` markers from the
         # stream; use them at [DONE] to flag is_cited on retrieved chunks.
         marker_stripper = CitationMarkerStripper()
@@ -227,6 +232,7 @@ async def create_message(
                             else _extract_text_from_sse(full_response)
                         )
                         if not _is_refusal(final_text):
+                            persisted_sources = source_citations
                             sources_json = json.dumps(source_citations)
                             yield f"event: sources\ndata: {sources_json}\n\n"
                     full_response.append(sse_chunk)
@@ -256,22 +262,12 @@ async def create_message(
             # can exit cleanly.
             assistant_text = _extract_text_from_sse(full_response)
             if assistant_text:
-                # Apply the same refusal detection used for the live SSE
-                # `event: sources` suppression so reloading the conversation
-                # later doesn't bring the misleading chip back. Prefer the
-                # final-round text when available — it's what the SSE path
-                # checks — and fall back to the full reconstructed text if
-                # the final_text buffer wasn't populated (e.g. pre-tool
-                # flow). If the message is a refusal we persist sources=None
-                # regardless of what the tool calls retrieved; the frontend
-                # renders the chip based on the stored field, so dropping
-                # it here keeps the reload UX consistent with the stream.
-                refusal_check_text = final_text_buf[0] if final_text_buf else assistant_text
-                sources_to_persist: list[dict] | None = (
-                    None
-                    if not source_citations or _is_refusal(refusal_check_text)
-                    else source_citations
-                )
+                # Persist exactly the sources that were emitted live (issue #277).
+                # `persisted_sources` is None when nothing was emitted: refusal,
+                # no chunks, or an interrupted stream that never reached [DONE].
+                # Persisting this verbatim — no second refusal check, no
+                # re-derivation — keeps reload identical to the live view.
+                sources_to_persist: list[dict] | None = persisted_sources
                 try:
                     await asyncio.shield(
                         repository.create_message(
--- a/app/backend/tests/test_sources_event.py
+++ b/app/backend/tests/test_sources_event.py
@@ -1152,3 +1152,280 @@ class TestChunkExpansionIntegration:
         assert "hello world snippet" in output
         assert "Test Video" in output
         assert "data: [DONE]" in output
+
+
+class TestSourcesPersistMatchesLive:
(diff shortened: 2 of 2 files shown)


coleam00:
**Benchmark Evaluator** — score: **53/70**

A clean, disciplined, faithfully-planned change that establishes a sound single-source-of-truth for assistant-message sources: the value emitted in the live `event: sources` SSE is captured once and persisted verbatim, eliminating a latent double-decision between the `[DONE]` branch and the `finally` block. Scope, code quality, and plan fidelity are excellent, and the subtle handling (capture before yield, finalized list, type correctness) is thoughtful.

The weakness is in actual bug-fixing impact and test rigor. In the current code `source_citations` is only populated inside the `[DONE]` branch, so an interrupted stream that never reaches `[DONE]` already persisted `sources=None` under the old logic — the literal reported symptom (duplicate/longer chips after interruption) doesn't reproduce on the path this fix touches. Consequently neither regression test fails on the pre-fix code; they characterize and lock the invariant rather than prove the reported bug was caught and fixed. The change is a good defensive refactor and a real robustness improvement, but it's stronger as a contract-hardening than as a demonstrated root-cause repair.

## Scores

|---|---|---|
| solves_issue | 6/10 | Enforces the invariant the issue explicitly demands — persisted sources equal the live `event: sources` payload by construction. But the specific reported symptom (duplicate/longer chips after an interruption) doesn't map to a code path this changes: `source_citations` is only populated inside the `[DONE]` branch, so an interrupted stream that skips `[DONE]` already persisted `None` under the old code. The fix hardens the contract more than it repairs a reproducing failure. |
| root_cause | 7/10 | Collapsing the live-emit and persist decisions into a single `persisted_sources` variable is the correct structural direction and removes a genuine latent double-decision (two independent `_is_refusal` calls). However, the exploration itself flagged that the divergence it eliminates may not be the actual source of the reported duplicates, so it reads as sound-but-preventive rather than a confirmed root-cause repair. |
| scope_discipline | 9/10 | Surgical: one production file with a ~6-line net change plus a single test class. Removes the now-dead `refusal_check_text` derivation and stale comment, touches nothing out of scope (no frontend, no repository, no footgun cleanups). |
| subtle_correctness | 8/10 | Captures the reference at the single emission point immediately before the yield, correctly anticipating client-disconnect/CancelledError after the bytes flush. `source_citations` is finalized (deduped/collapsed/capped) before capture and not mutated afterward, so the shared reference is safe. Types are correct (`list[dict] | None`). |
| code_quality | 8/10 | Clean, idiomatic, well-commented with issue references; the new comment clearly states why persistence mirrors the live emission. Matches surrounding style and comment density. Slight ding: the inline comments somewhat over-claim the bug behavior the code path actually exhibits. |
| test_discipline | 5/10 | Two well-structured ASGI integration tests with good mocks and explicit equality assertions. The real weakness: neither test fails on the pre-fix code — the interrupted case already persisted `None`, and the completed case already persisted the collapsed list — so they lock the invariant going forward but do not demonstrate the reported regression. The interrupted test's stated premise ('the finally block could independently re-derive a duplicate-laden list') isn't true given `source_citations` is empty on that path. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan essentially verbatim — same `persisted_sources` variable, same capture point before the yield, same `finally` simplification, and the two named test cases under the exact `TestSourcesPersistMatchesLive` class the plan specified. Zero drift. |

---

coleam00:

This is a clean, surgical, and well-reasoned refactor that establishes a single source of truth for streamed-vs-persisted sources, executed in faithful lockstep with an unusually precise plan. Code quality, scope discipline, and plan fidelity are excellent, and no new correctness hazards are introduced (the capture-before-yield detail is genuinely thoughtful). The structural change — eliminating the duplicated emit/persist decision — is the right long-term design and makes the issue's invariant hold by construction.

The weakness is evidentiary. In the current codebase source_citations is populated exclusively inside the [DONE] branch, so the two decision points already produced identical results in every reachable path, and an interrupted stream already persisted None. As a result the change is behavior-preserving rather than a demonstrable bug fix, and the two added 'regression' tests pass on both the old and new code — they lock in current behavior but do not prove the reported divergence existed or was fixed (and one docstring overstates the pre-fix behavior). A truly convincing PR would have included a test that fails before the change. Net: high-craft, well-scoped hardening whose correctness is sound but whose claimed root cause and regression coverage are not substantiated against this code.

## Scores

|---|---|---|
| solves_issue | 7/10 | The fix makes 'persisted sources == live-emitted sources' true by construction via a single bridge variable, which is exactly the guarantee the issue demands. Caveat: against the current code the two paths already agreed in every reachable case (source_citations is populated only inside the [DONE] branch, so an interrupted stream that never reaches [DONE] already persisted None), so this reads as a hardening refactor rather than the repair of an observably reproducing bug. |
| root_cause | 7/10 | Collapsing the duplicated emit/persist decisions into one source of truth is the structurally correct instinct and permanently forecloses future drift. Marked down because the stated root-cause mechanism (the finally block 're-deriving a different, longer, duplicate-laden list' on interrupt) cannot actually occur here: source_citations is empty when [DONE] is skipped, so the diagnosis is elegant but not tied to a real reproduction. |
| scope_discipline | 10/10 | Two files, ~6 net production lines plus tests — precisely the surface the plan scoped, with no collateral edits, no touching of protected/auth/repository code or the finalization pipeline. |
| subtle_correctness | 8/10 | persisted_sources is captured before the yield (CancelledError/disconnect-safe), None semantics are correct for refusal/empty/interrupted, the reference assignment is safe since the list isn't mutated after [DONE], and types (list[dict] | None) keep mypy clean. No foot-guns introduced. |
| code_quality | 9/10 | Clean, idiomatic, and well-commented; the new comments explain intent and reference the issue, and removing the stale re-derivation comment block leaves the finally block clearer than before. Matches surrounding style closely. |
| test_discipline | 5/10 | Two readable, well-constructed integration tests are added, but neither fails on the pre-fix code: the interrupted test asserts sources=None, which the old finally block already produced (empty source_citations), and the completed test asserts emitted==persisted, which the old code also satisfied. They are characterization tests dressed as regression tests, and a docstring claim ('Before the fix the finally block could independently re-derive a different, longer, duplicate-laden list') is not actually true for the interrupted path. |
| plan_impl_fidelity | 10/10 | The diff executes the plan almost verbatim — same persisted_sources variable name, identical placement and comments, capture-before-yield at the single emission point, removal of the refusal_check_text re-derivation, and both planned test cases under the named TestSourcesPersistMatchesLive class. |

---

PR #288, opened 2026-05-31
[bench-OK] review: add missing refusal test for issue #277

coleam00:
Fixes #277

**Source run-id:** `300ac0f4-f7c9-4946-a908-e7be0e5fe888`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -171,6 +171,9 @@ async def create_message(
     async def event_generator() -> AsyncGenerator[str, None]:
         full_response: list[str] = []
         final_text_buf: list[str] = []
+        # Single source of truth for what was actually emitted on the wire.
+        # None means the `event: sources` frame was never delivered.
+        sources_finalized: list[dict] | None = None
         # Two-tier citations (issue #176): strip `[c:<id>]` markers from the
         # stream; use them at [DONE] to flag is_cited on retrieved chunks.
         marker_stripper = CitationMarkerStripper()
@@ -229,6 +232,7 @@ async def create_message(
                         if not _is_refusal(final_text):
                             sources_json = json.dumps(source_citations)
                             yield f"event: sources\ndata: {sources_json}\n\n"
+                            sources_finalized = source_citations
                     full_response.append(sse_chunk)
                     yield sse_chunk
                     continue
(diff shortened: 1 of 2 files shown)


coleam00:

A strong, structurally-correct fix. It identifies the real defect — persistence in the `finally` block independently re-derived sources via a second `_is_refusal` call over a possibly-mutated object, so it could disagree with the live `event: sources` frame, and on client disconnect would persist chips the user never saw. The fix collapses both decisions into a single `sources_finalized` variable bound only after the frame is actually yielded, guaranteeing emitted == persisted across clean, refusal, no-sources, errored, and disconnected paths. Scope is tight (two files, exactly as planned) and the change removes dead code rather than adding complexity.

The test suite is the standout: four targeted regression tests, including a disconnect test that closes the generator while suspended at the `event: sources` yield to prove the post-yield assignment never runs — directly validating the mechanism the fix relies on. Plan-to-implementation fidelity is near-perfect, and the commit split shows the self-review step correctly catching and adding the refusal test the implementation initially omitted. Minor deductions only for inherent reliance on subtle generator-cleanup-time await semantics, which is an established pattern in this codebase but remains delicate.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes the live-vs-reload divergence on interrupted/errored answers by making persisted sources a faithful echo of what was emitted. Legacy stale rows are correctly left out of scope, so it solves the going-forward code path fully. |
| root_cause | 10/10 | True root cause: it eliminates the second, independent refusal/source re-derivation in `finally` and replaces it with one `sources_finalized` source of truth set only after the frame is on the wire. Two decision sites that could disagree become one. |
| scope_discipline | 10/10 | Surgical — only `messages.py` and the existing `test_sources_event.py` are touched, exactly the two files the plan named. No drift into auth/rate-limit/frontend. |
| subtle_correctness | 9/10 | Correctly exploits generator suspension semantics — placing the assignment after `yield` means a disconnect/GeneratorExit at that point leaves `sources_finalized=None`. Async shield, refusal, no-sources, and error-before-DONE paths all converge correctly. No foot-guns introduced. |
| code_quality | 9/10 | Clean and idiomatic: removes a dead `refusal_check_text` local and a stale comment block, adds a concise single-source-of-truth comment. Reads like the surrounding code. |
| test_discipline | 9/10 | Four meaningful regression tests covering error-before-DONE, disconnect-at-sources-yield (via real `aclose()`), refusal, and clean multi-chunk collapse — each asserting live SSE == persisted `sources`. The disconnect test genuinely exercises the fix's crux. Self-review commit added the refusal test the impl initially missed. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan almost verbatim: the `sources_finalized` init, post-yield assignment, `finally` direct read, and removal of the re-derivation all match, and all four planned tests are present. |

---

PR #289, opened 2026-05-31
[bench-KO] impl: finalize sources consistently after interrupted responses

coleam00:
Fixes #277

**Source run-id:** `dd3a065f-e0b9-4db8-b98e-550db60f3539`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -122,7 +122,6 @@ async def create_message(
     # lists exactly what the model actually read. The video_id whitelist is
     # only consulted by the transcript tool (it guards against hallucinated
     # ids); the search tools ignore it.
-    source_citations: list[dict] = []
     tool_chunks_acc: list[dict] = []
     embedding_cache: dict[str, list[float]] = {}
     tools_param: list[dict] | None = None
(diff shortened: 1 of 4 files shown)


coleam00:

A strong, well-reasoned PR. It correctly identifies the dual root cause — the backend's fragile in-place mutation of a shared `source_citations` list, and the frontend's `onComplete` firing even on mid-stream error — and addresses both with a clean structural fix. Extracting `_finalize_sources` as the single deterministic source of truth (recomputed from raw tool chunks, with copied dicts) is exactly the right move and is an improvement over the prior inlined mutation. Scope is tight, code quality is high, and the documentation ties each pipeline step back to its originating issue.

Weaknesses are minor. The regression test for the post-[DONE] interrupted path was softened from the plan's disconnect simulation to a clean-completion equality assertion, so the actual interrupt-after-emit persistence path is verified only indirectly. There is a small refusal-text fallback asymmetry between the live emit ("") and persist (assistant_text), and the two frontend changes overlap — with the onComplete guard in place, the ChatArea ghost-cleanup backstop only matters for a narrow post-onComplete throw. None of these are correctness defects; they are polish gaps. Overall a high-fidelity, root-cause fix.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes the live-vs-reload mismatch on both sides: backend routes persistence through the same finalization the live SSE uses, and the frontend stops injecting a ghost assistant bubble on mid-stream error. Persisted sources are now deduped/collapsed/capped/refusal-checked identically to what was shown. |
| root_cause | 9/10 | True structural fix. It eliminates the fragile shared-mutable-list architecture by extracting a pure `_finalize_sources` helper that recomputes from raw tool chunks each time, so the emit and persist paths cannot diverge — not a band-aid on the symptom. |
| scope_discipline | 9/10 | Surgical: 4 files, every one necessary (backend handler, frontend hook, frontend component, test file). Touching the protected `messages.py` is unavoidable since the source pipeline lives there, and the change is squarely on-issue rather than auth/rate-limit logic. |
| subtle_correctness | 8/10 | Thoughtful: copies dicts before setting is_cited to avoid mutating the accumulator, preserves the asyncio.shield save, clean None semantics. Minor asymmetry — the live emit uses "" as refusal-text fallback while the persist falls back to assistant_text — and the frontend changes are partly redundant (the onComplete guard makes the ChatArea backstop a no-op on genuine errors), but no real bugs introduced. |
| code_quality | 9/10 | Production-grade: thorough docstring documenting the pipeline and the issues it satisfies (#277/#208/#176/#158), idiomatic Python, clear comments on the frontend rationale. Reads consistently with surrounding code. |
| test_discipline | 8/10 | Two genuine ASGI integration tests: one asserts sources=None and no sources event on mid-stream error, one asserts persisted == live-emitted with collapse-by-video. Meaningful and targeted, but the planned 'disconnect after [DONE]' (aclose) scenario was softened to a clean-completion equality check, so the interrupted-after-DONE persist path isn't directly exercised. |
| plan_impl_fidelity | 9/10 | Faithfully executes the plan: `_finalize_sources` extraction with both call sites, the `if (!streamError) onComplete` guard, the `pendingAssistantMsgIdRef` cleanup, and the regression test class all match. Only drift is test (b) being renamed and implemented as a clean-completion equality check rather than simulating a post-[DONE] disconnect via aclose(). |

---

PR #290, opened 2026-05-31

### LINKED ISSUE #278: Shareable read-only conversation links
As a user who just had a great Q&A with the video assistant, I want to **share that conversation** with someone who doesn't have an account.

I'd like to generate a link I can send them. Anyone with the link should see a clean, read-only view of the full exchange: my questions, the assistant's answers, and the video citations. But they should **not** be able to send new messages, see any of my other conversations, or do anything as me. People I share with shouldn't need to log in to read it.

I also want to be able to turn a link off later so it stops working.

Done looks like: I can create a share link for a conversation, open it in a logged-out browser and read the whole conversation read-only, and revoke it so it no longer opens.

coleam00:
Fixes #278

**Source run-id:** `a2e79440-de6f-4e4c-a9f7-3dd4cfd0c957`


coleam00:
Dropped from benchmark: issue #278 violates MISSION Hard Invariant #3 (conversations are private; no share links). 3/4 cells correctly refused; recorded as a governance-compliance side-finding.

Issue #279, opened 2026-05-31
Scope a conversation to specific videos


coleam00:
As a user, sometimes I want to dig into one or a few specific videos rather than the whole library.

I want to be able to pick which videos a conversation should draw from, so that the assistant only answers using those videos and its citations only come from them. Once I've scoped a conversation to certain videos, that scope should stick for the rest of that conversation. If I don't pick anything, it should search across everything like it does today.

Done looks like: I can choose a subset of videos for a conversation, and from then on the assistant's answers and citations in that conversation only come from the videos I picked.

PR #296, opened 2026-05-31
[bench-OK] review: add missing tests and fix broken stubs for video scope (issue #279)

coleam00:
Fixes #279

**Source run-id:** `888cd9ce-ea5b-4d3f-9439-4d9d46e8d038`

### DIFF
--- a/app/backend/alembic/versions/0006_add_video_filter_to_conversations.py
+++ b/app/backend/alembic/versions/0006_add_video_filter_to_conversations.py
@@ -1,0 +1,33 @@
+"""Add video_filter JSONB to conversations.
+
+Stores an optional array of video_id strings that scopes retrieval
+for the lifetime of a conversation. NULL or omitted = search all videos.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        ALTER TABLE conversations
+        ADD COLUMN IF NOT EXISTS video_filter JSONB
+        """
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE conversations DROP COLUMN IF EXISTS video_filter")
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -253,6 +253,7 @@ async def keyword_search(
     top_k: int,
     language: str = "english",
     allowed_source_types: list[str] | None = None,
+    video_ids: list[str] | None = None,
 ) -> list[dict]:
     """
     Return top-K chunks matching a full-text query using tsvector.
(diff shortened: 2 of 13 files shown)


coleam00:

A high-quality, well-scoped implementation of conversation-level video scoping. It chooses the right architecture — a persistent nullable JSONB column read server-side each turn, plus an optional video_ids ACL threaded through the existing retrieval stack as an intersecting SQL clause — so the scope cannot be escaped by search or transcript tools, and the source-type ACL still wins for non-members. The diff is surgical, tracks the Opus plan almost exactly, and the Sonnet self-review pass added genuinely meaningful backend tests rather than gaming existing ones. Weaknesses are minor: no frontend tests were added despite being planned, the ACL-intersection guarantee is verified only at the SQL-construction level instead of end-to-end, ChatArea leans on verbose inline styles, and a failed whitelist load combined with an active filter leaves the transcript tool unguarded (search remains scoped, so impact is small). Overall a strong, near-faithful execution of the plan that solves the issue at its root.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers the full feature: user picks a subset of videos at conversation creation, the scope is stored server-side as video_filter and read on every turn, search results and citations are constrained to those videos, and empty/no selection searches everything. Matches the issue's 'done looks like' exactly; scope is fixed-at-creation per the issue's 'sticks for the rest of the conversation'. |
| root_cause | 9/10 | Structurally sound, not a band-aid. Persists scope as a nullable JSONB column (NULL/empty = all), threads an optional video_ids ACL through tools -> retriever -> SQL as an additional AND clause that intersects with (never replaces) the source_type ACL, and narrows the transcript whitelist. Mirrors the established allowed_source_types pattern. |
| scope_discipline | 9/10 | Surgical: every one of the 13 files maps to a step in the plan with no unrelated edits. Protected files (conversations.py, messages.py) are touched but only minimally and only as the feature genuinely requires. |
| subtle_correctness | 8/10 | Thoughtful: empty-list normalization to None, AND-intersection guarding non-members from dynamous chunks, is_member captured at turn start, whitelist narrowing operates on a guaranteed-set value so no None-iteration crash. Minor: get_conversation assumes asyncpg returns JSONB as a string (consistent with existing sources handling), and a failed whitelist load + active filter leaves the transcript tool unguarded though search stays scoped — acceptable degradation. |
| code_quality | 8/10 | Backend is clean and idiomatic; the explicit if/else SQL variants are verbose but deliberately chosen to keep parameterization safe. VideoExplorer reuses Tailwind well. Dings: ChatArea adds heavy inline style objects (CLAUDE.md prefers Tailwind, though the file already uses inline styles) and a pointless handleToggleSelect passthrough wrapper. |
| test_discipline | 8/10 | Strong, targeted backend tests: SQL-clause construction with/without video_ids, ACL-intersection preservation, video_ids forwarding through execute_tool, and JSONB round-trip/empty-list normalization. Stub signature updates are legitimate (real signatures changed), not gaming. Gap: no frontend Vitest tests for createConversation/VideoExplorer despite the plan calling for them, and the ACL regression is asserted at SQL-construction level rather than true end-to-end zero-chunks. |
| plan_impl_fidelity | 9/10 | Implementation faithfully executes the plan nearly step-for-step: migration 0006, repository create/get/list + search filters, retrieve_hybrid forwarding, tools.py threading, conversations.py + messages.py narrowing, api.ts, VideoExplorer selection mode, ChatArea picker + scope chip, and the optional Sidebar badge. Only drift: the planned frontend tests are absent and the ACL regression is weaker than the planned end-to-end check. |

---

PR #297, opened 2026-05-31

coleam00:
Fixes #279

**Source run-id:** `c0eee35d-cad5-487d-a1f0-ba2e6eeb1f28`

### DIFF
--- a/app/backend/alembic/versions/0006_add_conversation_video_scope.py
+++ b/app/backend/alembic/versions/0006_add_conversation_video_scope.py
@@ -1,0 +1,35 @@
+"""Add scoped_video_ids to conversations.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        ALTER TABLE conversations
+        ADD COLUMN IF NOT EXISTS scoped_video_ids TEXT[]
+        """
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        """
+        ALTER TABLE conversations
+        DROP COLUMN IF EXISTS scoped_video_ids
+        """
+    )
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -253,6 +253,7 @@ async def keyword_search(
     top_k: int,
     language: str = "english",
     allowed_source_types: list[str] | None = None,
+    allowed_video_ids: list[str] | None = None,
 ) -> list[dict]:
     """
     Return top-K chunks matching a full-text query using tsvector.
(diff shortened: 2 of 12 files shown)


coleam00:

A strong, well-targeted implementation of conversation video scoping. It picks the correct architectural seam — extending the existing video_id_whitelist into the keyword and vector search SQL — so that scoped answers and citations are enforced at the retrieval layer, and it leans on SELECT * so persistence and read-back work without a missed query edit. Backend code is idiomatic, SQL is parameterized, empty-selection fallback is handled consistently end-to-end, and tests cover the repository, tool-dispatch, and frontend surfaces. Plan fidelity is high with only cosmetic drift (revision number, useMessages vs useConversations).

Weaknesses are modest: the new frontend modal and button are built entirely with inline style objects, violating the repo's explicit Tailwind-only convention; the most important integration test (that messages.py builds the whitelist from scoped_video_ids) was planned but not written; and the unscoped path needlessly passes the full video-id list as a filter. The PR also touches three protected-path files — unavoidable for this feature, but it would trip the factory's human-authored-only gate. None of these undermine correctness; the feature works as the issue describes.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers the full feature: a picker sets scoped_video_ids at creation, the whitelist is threaded into all three search tools so both answers and citations are restricted, the column persists and is re-read on every message so scope sticks, and an empty/absent selection falls back to searching everything. The transcript tool already honored the same whitelist, so scoping is consistent across retrieval paths. |
| root_cause | 9/10 | Structurally sound: it extends the existing video_id_whitelist mechanism (which the baseline only used as a transcript-tool guard) down into the actual retrieval SQL, rather than bolting on a separate filter. Restricting at the keyword/vector query level is the correct layer for the citations-only-from-scoped requirement. |
| scope_discipline | 8/10 | 12 files, all genuinely required (migration, repository, both routes, retriever, tools, frontend api/hook/component, tests); no unrelated drive-by changes. Does cross protected-path files (messages.py, conversations.py, repository.py), but only the necessary non-security surfaces and the feature is impossible to build without them. |
| subtle_correctness | 8/10 | Empty-list-to-None coercion is consistent across repository, messages.py, and the frontend; parameterized $4 SQL; TEXT[] maps cleanly to/from asyncpg lists; SELECT * means the new column propagates without a missed query edit. Minor: the unscoped path now passes the entire video-id list as an ANY() filter on every search (harmless but wasteful) rather than skipping the clause. |
| code_quality | 7/10 | Backend is clean and idiomatic — matches the existing conditional-SQL-append pattern and naming. The new frontend components, however, use ~150 lines of inline style={{}} objects, directly against CLAUDE.md's 'Tailwind utility classes only, no inline style' rule. |
| test_discipline | 7/10 | Meaningful, non-gamed tests at the repository boundary (allowed_video_ids SQL), the tool dispatcher (whitelist forwarding to all 3 search executors), persistence, and the frontend picker/indicator/API-payload. Gap: the planned test_message_handler_uses_scoped_whitelist — the key integration test that messages.py actually derives the whitelist from scoped_video_ids — was not written. |
| plan_impl_fidelity | 8/10 | Implementation faithfully follows the plan across all backend/frontend touchpoints; it correctly adapted the migration revision (0002 in plan -> 0006 chained off the real head) and satisfied the get/list SELECT requirement implicitly via SELECT *. Minor drift: edited useMessages.ts instead of the planned useConversations.ts, and omitted one of the four planned backend tests. |

---

PR #298, opened 2026-05-31
[bench-KO] impl: scope a conversation to specific videos (#279)

coleam00:
Fixes #279

**Source run-id:** `0c157926-d88b-4eac-82ff-20c90106a2a1`

### DIFF
--- a/app/backend/alembic/versions/0006_add_scoped_video_ids.py
+++ b/app/backend/alembic/versions/0006_add_scoped_video_ids.py
@@ -1,0 +1,37 @@
+"""Add scoped_video_ids column to conversations (issue #279).
+
+Lets a conversation be restricted to a user-selected subset of videos. When
+`scoped_video_ids` is non-null, retrieval (keyword + vector) only sees chunks
+whose `video_id` is in the list; NULL preserves the existing "search
+everything" behavior. Stored as TEXT[] because video ids are plain text UUIDs.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    # Nullable with no default — NULL means "no scope set" (search everything),
+    # which is the correct in-place upgrade for every existing conversation.
+    op.execute(
+        """
+        ALTER TABLE conversations
+        ADD COLUMN IF NOT EXISTS scoped_video_ids TEXT[]
+        """
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE conversations DROP COLUMN IF EXISTS scoped_video_ids")
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -253,6 +253,7 @@ async def keyword_search(
     top_k: int,
     language: str = "english",
     allowed_source_types: list[str] | None = None,
+    allowed_video_ids: list[str] | None = None,
 ) -> list[dict]:
     """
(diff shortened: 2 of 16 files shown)


coleam00:

A high-quality, well-scoped implementation of conversation-level video scoping. The standout strength is architectural correctness: the scope filter is applied at the retrieval SQL layer and threaded consistently through the entire RAG pipeline (messages route -> tools dispatcher -> hybrid retriever -> keyword/vector search), so both answers and citations are guaranteed to come only from the selected videos, with NULL preserving today's search-everything behavior. The code is clean, idiomatic to the repo's conventions (SQL confined to repository.py, asyncpg placeholders, Tailwind/inline-style parity with existing components), and thoroughly documented with issue references. It tracks the plan almost item-for-item.

Weaknesses are modest. The PR necessarily edits protected, auth-adjacent files (conversations/messages routes and repository owner-scoping), which the factory rules earmark for human authoring — the edits preserve owner scoping and are unavoidable for the feature, but they push the change outside the strictly-safe lane. The tests, while broad, verify SQL construction and pass-through wiring against a mocked DB rather than the real end-to-end behavioral assertion the plan called for (only scoped chunks returned), and the empty-list-yields-no-results semantic is a small latent API foot-gun. None of these undermine the feature; they keep an otherwise excellent PR just short of perfect.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully delivers the feature: a conversation can be scoped to a subset of videos, the scope persists on the conversation row (so it sticks for the rest of the conversation), and retrieval + citations are filtered to those video_ids at the SQL layer. NULL/no-selection preserves the search-everything default exactly as asked. |
| root_cause | 9/10 | Structurally sound. The filter is pushed down to the actual retrieval SQL (keyword_search + vector_search_pg) and threaded cleanly from messages route -> tools -> retriever, so there is no way for a scoped conversation to surface out-of-scope chunks or citations. Persisted as a column rather than a session hack, which is the correct place for conversation-level state. |
| scope_discipline | 8/10 | Changes map almost 1:1 to the plan's file list with no unrelated edits. The one caveat is that it necessarily touches protected/auth-adjacent paths (routes/conversations.py, routes/messages.py, repository owner-scoping) which the factory rules flag for human authoring — but the edits are additive, minimal, and preserve user_id owner scoping, so the touch is unavoidable for this feature. |
| subtle_correctness | 8/10 | Careful with edge cases: the ($4::text[] IS NULL OR video_id = ANY($4)) guard short-circuits cleanly and keeps the HNSW index usable; empty-list-vs-null semantics are documented; frontend sends null (not []) when nothing is selected. get_conversation uses SELECT * so the new column flows through without an explicit change. Minor latent foot-gun: a deliberately-passed empty list yields zero results, which is documented but easy to trip over via the API. |
| code_quality | 9/10 | Readable, idiomatic, well-commented with issue references and clear docstrings. Backend follows the SQL-in-repository convention and asyncpg $N placeholders; the modal uses Tailwind while the header button uses inline styles consistent with the existing Export button. Naming and structure match the surrounding codebase. |
| test_discipline | 8/10 | Strong coverage: repository SQL/param-binding, tool forwarding, retriever fan-out, and frontend modal interactions all tested, and existing-test signature updates are legitimate (not gaming). Slightly below the plan, which asked for an end-to-end test asserting only vid-a chunks return with real inserted chunks; the actual tests assert SQL construction and wiring against a mocked DB instead of real query behavior — reasonable given no Postgres in the env, but it is verification of plumbing rather than results. |
| plan_impl_fidelity | 9/10 | The diff faithfully executes nearly every plan item across migration, repository, retriever, tools, both routes, and all frontend pieces. Minor benign deviations: the migration uses op.execute(ALTER TABLE ... ADD COLUMN IF NOT EXISTS) instead of op.add_column/sa.Column (functionally equivalent, more idempotent); get_conversation wasn't explicitly edited (SELECT * covers it); and tests assert wiring rather than the planned real-data behavioral test. |

---

Issue #280, opened 2026-05-31
Regenerate the last assistant response


coleam00:
As a user, sometimes the assistant's answer misses the mark or stops short.

I want a "regenerate" option on the most recent answer so I can get a fresh take on my last question without retyping it. The new answer should replace the old one in the conversation, come with its own citations, and the streaming experience should feel just like sending a normal message.

Regenerating shouldn't let me sneak past my daily message limit.

Done looks like: I can click regenerate on the latest answer, watch a new answer stream in to replace it (with its own sources), and the action is counted fairly against my usage.

PR #300, opened 2026-05-31
[bench-OO] review: fix mock parameter names in test_regenerate.py

coleam00:
Fixes #280

**Source run-id:** `a70cd8a1-c0a9-4c0e-8111-5ef94dd9ef42`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -603,6 +603,25 @@ async def list_messages(conversation_id: str, user_id: str) -> list[dict]:
     return results
 
 
+async def delete_message(message_id: str, user_id: str) -> bool:
+    """Delete a single message scoped to its owner. Returns True if a row was deleted.
+
+    Owner-scoped via JOIN on conversations.user_id so a caller cannot delete
+    another user's message even if it forgets to check (mirrors create_message).
+    """
+    async with _acquire() as conn:
+        result = await conn.execute(
+            """
+            DELETE FROM messages m
+            USING conversations c
+            WHERE m.id = $1 AND m.conversation_id = c.id AND c.user_id = $2
+            """,
+            message_id,
+            user_id,
+        )
+    return result != "DELETE 0"  # type: ignore[no-any-return]
+
+
 # ---------------------------------------------------------------------------
 # Channel sync runs
 # ---------------------------------------------------------------------------
--- a/app/backend/main.py
+++ b/app/backend/main.py
@@ -103,7 +103,15 @@ app.add_middleware(
 # ---------------------------------------------------------------------------
 # Routes (imported here to keep main.py clean)
 # ---------------------------------------------------------------------------
-from backend.routes import admin, auth, channels, conversations, ingest, messages  # noqa: E402
+from backend.routes import (  # noqa: E402
+    admin,
+    auth,
+    channels,
+    conversations,
+    ingest,
+    messages,
+    regenerate,
+)
 
 # Auth routes are public (signup/login don't require a session; /me and /logout
 # rely on their own dependency/cookie behaviour).
@@ -114,6 +122,7 @@ app.include_router(auth.router, prefix="/api")
 _auth_required = [Depends(get_current_user)]
 app.include_router(conversations.router, prefix="/api", dependencies=_auth_required)
 app.include_router(messages.router, prefix="/api", dependencies=_auth_required)
+app.include_router(regenerate.router, prefix="/api", dependencies=_auth_required)
 
 # Library-mutation routes (ingest a video, backfill the whole channel) and
 # admin routes — all gated on get_current_admin. These endpoints write to the
--- a/app/backend/routes/regenerate.py
+++ b/app/backend/routes/regenerate.py
@@ -1,0 +1,290 @@
+"""
+Regenerate route — POST /api/conversations/{conv_id}/regenerate
+
+Re-runs the tool-driven RAG flow for the trailing user question and replaces
+the most recent assistant answer with a fresh one (its own citations). Mirrors
+the streaming/persistence idiom of ``routes/messages.py`` exactly:
+
+  1. Verify conversation ownership (404 cross-user, no leak)
+  2. Guard: there must be a trailing assistant message to regenerate (409)
+  3. Enforce the 25 msg/user/24h cap (same call as the send path — regenerate
+     cannot bypass the limit)
+  4. Re-stream the response against the existing history (minus the stale
+     answer)
+  5. Send the sources event (the model's tool calls) before [DONE]
+  6. Persist the new assistant message, THEN delete the old one — so a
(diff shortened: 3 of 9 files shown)


coleam00:

A strong, disciplined feature PR (OO cell, plan=Opus/impl=Opus). It fully solves issue #280: a Regenerate affordance on the latest assistant answer that re-streams a fresh, separately-cited reply, replaces the stale one, and is correctly billed against the daily cap. The architecture is the right one given the governance constraints — work is isolated in a new non-protected route module that reuses messages.py helpers read-only, with only additive touches to the two protected files, exactly as planned. Subtle correctness is well handled: 409/404/429 ordering preserves the cap and the old answer on every failure path, and shielded persist-then-delete avoids orphaned questions. Test coverage is meaningful and targeted across backend route, backend repository, and frontend layers, and the self-review commit demonstrates the suite was run. The only real blemishes are minor: a redundant double-computation of is_member, and a Regenerate button styled with inline CSS and JS hover handlers that bends CLAUDE.md's Tailwind-only convention (albeit to match existing buttons). Plan-implementation fidelity is essentially exact. Net: high-quality, ships-ready work.

## Scores

|---|---|---|
| solves_issue | 9/10 | Hits every 'done looks like' bullet: a Regenerate button on the latest assistant message, a fresh streamed replacement with its own citations (event: sources), the old answer replaced, and the action billed against the 25/24h cap server-side. Only short of 10 because correctness rests on mocks, not a live end-to-end run. |
| root_cause | 9/10 | Structurally sound feature work rather than a band-aid: a dedicated non-protected route module reuses the proven messages.py streaming/persistence idiom, adds an owner-scoped DELETE, and orders persist-new-then-delete-old so a failed regenerate never orphans a question. The rate-limit call is the same one the send path uses, so the cap genuinely cannot be bypassed. |
| scope_discipline | 9/10 | Nine files, all directly required; no drive-by refactors or unrelated 'improvements'. The two protected-file touches (main.py router line, repository.delete_message) are strictly additive and were explicitly flagged in the plan; the streaming hook refactor is a clean shared-helper extraction, not scope creep. |
| subtle_correctness | 8/10 | Thoughtful about failure modes: 409-guard before rate-limit so no-ops don't consume the cap, 404 (not 403) to avoid leaking foreign conversations, asyncio.shield on both DB writes, delete only after a successful persist, and is_member captured once per turn. Minor smell: is_member is computed twice (is_member_for_turn then turn_is_member) which is redundant but harmless; no actual bug spotted. |
| code_quality | 8/10 | Backend is production-grade — clear naming, dense explanatory comments, faithful to messages.py conventions and the asyncpg/SQL rules. The Message.tsx button uses heavy inline styles plus onMouseEnter/Leave hover handlers, which technically deviates from CLAUDE.md's 'Tailwind only' rule, though it matches the existing ghost-button pattern in the codebase. |
| test_discipline | 9/10 | Genuinely targeted coverage: 5 route tests (happy path replace+count, own-sources, 429-preserves-old, 409-no-cap, 404), a parametrized repository unit test asserting owner-scoped parameter order, plus frontend Message and useStreamingResponse cases. The self-review commit fixing mock parameter names shows the tests were actually exercised, not gamed. |
| plan_impl_fidelity | 10/10 | Near-perfect execution of the plan — the route structure, the delete_message body, the consumeStream extraction, the ChatArea optimistic-restore handler, and every named test all match the plan almost line-for-line, including the documented persist-then-delete ordering and the protected-file caveats. |

---

coleam00:

An excellent, high-fidelity implementation of issue #280. The feature is complete and correct: a Regenerate affordance on the latest assistant message that re-runs the real tool-driven RAG flow, streams a replacement with its own citations, and is fairly metered against the daily cap. The engineering shows genuine care for subtle failure modes — rate-limit and emptiness guards ordered so a no-op never consumes the cap, persist-new-before-delete-old so an aborted regenerate never orphans the user's question, shielded DB writes against client-disconnect cancellation, and a frontend optimistic-remove-with-restore-on-failure that mirrors the existing send path. Tests are comprehensive and meaningful across backend route, repository, and frontend layers, mocked at the correct boundaries.

Weaknesses are minor. The PR additively touches two protected files (main.py, repository.py); this is genuinely necessary for the feature and the plan explicitly surfaced the governance caveat, but a strict file-granularity validator would flag it for human escalation. The create-succeeds/delete-fails edge can momentarily leave a duplicate assistant message (preferred over an orphan, and logged). The Regenerate button leans on large inline style objects instead of Tailwind utilities, a small deviation from the frontend conventions. None of these undermine correctness. This is among the stronger cells one would expect from an Opus-plans/Opus-implements pairing.

## Scores

|---|---|---|
| solves_issue | 10/10 | Delivers exactly what #280 asked: a Regenerate button on the latest assistant message that re-streams a fresh answer with its own citations, replaces the old one, and counts against the 25/24h cap via the same check_and_record call the send path uses. |
| root_cause | 9/10 | Structurally sound: a dedicated owner-scoped delete_message, persist-new-then-delete-old ordering so an aborted regen never orphans the user turn, and rate-limit enforced before any work. Reuses the real RAG/streaming pipeline rather than faking it. Minor: a create-succeeds/delete-fails path can leave a duplicate assistant message (logged, swallowed). |
| scope_discipline | 8/10 | Surgical — every changed file is load-bearing for the feature. It does touch two protected files (main.py, repository.py), but only additively (one include_router line, one new function), and the plan explicitly flagged this governance caveat rather than silently editing existing auth/scoping logic. |
| subtle_correctness | 9/10 | Careful about the hard parts: asyncio.shield around DB writes vs client-disconnect CancelledError, 409 guard ordered before rate-limit so a no-op never burns the cap, 404-not-403 to avoid leaking foreign conversations, refusal-suppressed sources, marker-stripper flush, frontend optimistic-remove-with-restore-on-failure. The turn_is_member/is_member_for_turn duplication is redundant but harmless. |
| code_quality | 9/10 | Production-grade and idiomatic: faithfully mirrors messages.py conventions, reuses its pure helpers read-only, thorough docstrings explaining the why. The new shared consumeStream helper cleanly deduplicates the fetch/SSE loop. Minor nit: the Regenerate button uses large inline style objects rather than Tailwind utilities, slightly against the frontend styling convention. |
| test_discipline | 10/10 | Strong targeted coverage at every boundary: backend happy-path (replace+count), own-sources, 429-preserves-old, 409-no-cap, 404-foreign, a parametrized repository unit test asserting owner-scoped DELETE arg order, plus frontend Message button render/click/streaming/role cases and a hook test for the regenerate POST and its 429→RateLimitError path. Mocked at the LLM/tool boundary per CLAUDE.md rules. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — file set, route step ordering, delete_message signature/SQL, the consumeStream extraction, Message/ChatArea wiring, and all five backend + repository + two frontend test cases all match the plan exactly, down to the 404/409 ordering and the persist-then-delete invariant. |

---

PR #301, opened 2026-05-31
[bench-KK] review: add missing tests for regenerate endpoint

coleam00:
Fixes #280

**Source run-id:** `aa05d60d-37f6-42d4-b906-d5f3b980cb7f`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -577,6 +577,43 @@ async def create_message(
     }
 
 
+async def get_message(message_id: str, user_id: str) -> dict | None:
+    """Return a single message only if its conversation belongs to the user."""
+    async with _acquire() as conn:
+        row = await conn.fetchrow(
+            """
+            SELECT m.*
+            FROM messages m
+            JOIN conversations c ON c.id = m.conversation_id
+            WHERE m.id = $1 AND c.user_id = $2
+            """,
+            message_id,
+            user_id,
+        )
+    if not row:
+        return None
+    d = dict(row)
+    d["sources"] = json.loads(d["sources"]) if d.get("sources") else None
+    return d
+
+
+async def delete_message(message_id: str, user_id: str) -> bool:
+    """Delete a message. Returns False if it does not belong to the user."""
+    async with _acquire() as conn:
+        result = await conn.execute(
+            """
+            DELETE FROM messages
+            USING conversations
+            WHERE messages.id = $1
+              AND conversations.id = messages.conversation_id
+              AND conversations.user_id = $2
+            """,
+            message_id,
+            user_id,
+        )
+    return result != "DELETE 0"  # type: ignore[no-any-return]
+
+
 async def list_messages(conversation_id: str, user_id: str) -> list[dict]:
     """Return messages only if the conversation belongs to the given user."""
     async with _acquire() as conn:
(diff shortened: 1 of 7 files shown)


coleam00:
**Benchmark Evaluator** — score: **53/70**

A competent, near-complete implementation of the regenerate feature that hits every requirement in the issue: a button on the final assistant message, a fresh streamed answer with its own citations replacing the old one, and a properly reused rate-limit check that prevents quota bypass. The backend is the strong half — well-scoped, ownership-guarded, and faithful to the existing create_message streaming/persistence pattern including asyncio.shield, with create-before-delete ordering that avoids data loss. Endpoint test coverage is meaningful, especially the assertion that the 429 path doesn't record against the cap.

The weaknesses are concentrated in the frontend and the test stubs. Message.tsx abandons the codebase's Tailwind-only convention for large inline style blocks and imperative hover mutation, and useStreamingResponse duplicates the entire SSE parsing loop rather than sharing it. On testing, the repository 'tests' are signature-inspection stubs that prove nothing about ownership enforcement, and the planned frontend tests are missing entirely, so the visible parts of the change (the UI button, the hook) go unverified. Worth noting for process: this touches repository.py message functions and messages.py, both protected paths per CLAUDE.md, which may warrant human authorship regardless of the score.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers a full regenerate feature: a button on the last assistant bubble, a new streaming answer that replaces the old one with its own sources, and a reused rate-limit check so it counts against the daily cap. Every 'done looks like' bullet from the issue is met. |
| root_cause | 8/10 | Structurally sound — a dedicated endpoint that re-runs the real RAG pipeline, persists the new message and deletes the old via asyncio.shield, and reuses rate_limit.check_and_record rather than faking a counter bump. delete+create instead of in-place update is a reasonable design given no update path existed. |
| scope_discipline | 9/10 | Touches exactly the seven files the work required (repository, messages route, hook, Message, ChatArea, api.ts, plus a test file) with no unrelated edits. Only nit is the regenerateMessage wrapper added to api.ts that the hook never actually calls (dead code). |
| subtle_correctness | 8/10 | Thoughtful about failure modes: ownership-scoped get/delete, create-before-delete order to avoid losing the old answer, asyncio.shield against client disconnect, refusal/citation handling mirrored from create_message, and all 404/403/400/429 branches guarded. Minor redundancy (is_member read twice) and a duplicate-message window if delete fails after create, but they re-raise. |
| code_quality | 6/10 | Backend is clean and idiomatic, closely mirroring create_message. Frontend lets it down: the regenerate button uses heavy inline style objects plus imperative onMouseEnter/Leave style mutation, directly violating the repo's Tailwind-only rule, and startRegenerate copy-pastes ~150 lines of SSE-parsing instead of sharing startStream's loop. |
| test_discipline | 6/10 | The endpoint tests are genuinely good — happy path asserts create/delete args, plus 404/403/400/400/429 and a check that the 429 path does NOT increment the counter. But the two repository 'tests' only inspect.signature for a user_id param (a gamed stub, not the real ownership-enforcement test the plan asked for), and the planned frontend component/hook tests are entirely absent. |
| plan_impl_fidelity | 7/10 | Backend implements the plan almost verbatim (only the delete/create order is swapped, harmlessly). Frontend follows the structure but drifts: startRegenerate uses a single onComplete callback rather than the planned onToken/onSources/onComplete/onError set, in-place replace became remove-then-append, and three planned tests (repository ownership, Message component, hook) were not delivered. |

---

PR #303, opened 2026-05-31

coleam00:
Fixes #280

**Source run-id:** `184805ef-3657-4573-a82b-89b460574da4`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -603,6 +603,22 @@ async def list_messages(conversation_id: str, user_id: str) -> list[dict]:
     return results
 
 
+async def delete_message(message_id: str, conversation_id: str, user_id: str) -> bool:
+    """Delete a message scoped to the conversation owner."""
+    async with _acquire() as conn:
+        result = await conn.execute(
+            """
+            DELETE FROM messages
+            WHERE id = $1 AND conversation_id = $2
+              AND EXISTS (SELECT 1 FROM conversations WHERE id = $2 AND user_id = $3)
+            """,
+            message_id,
+            conversation_id,
+            user_id,
+        )
+        return result != "DELETE 0"  # type: ignore[no-any-return]
+
+
 # ---------------------------------------------------------------------------
 # Channel sync runs
 # ---------------------------------------------------------------------------
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -54,68 +54,18 @@ class MessageCreate(BaseModel):
 
 
 # ---------------------------------------------------------------------------
-# POST /api/conversations/{conv_id}/messages
+# Shared streaming helper
 # ---------------------------------------------------------------------------
 
-
-@router.post("/conversations/{conv_id}/messages")
-async def create_message(
+async def _run_assistant_stream(
+    *,
     conv_id: str,
-    body: MessageCreate,
-    current_user: dict[str, Any] = Depends(get_current_user),
-):
-    """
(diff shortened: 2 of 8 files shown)


coleam00:

A strong, high-fidelity implementation of issue #280. The backend reuses the existing tool-driven streaming/persistence machinery via a clean shared helper instead of duplicating it, adds a correctly ordered regenerate endpoint (ownership -> latest-assistant guard -> preceding-user-turn guard -> rate-limit -> delete -> rebuild history -> stream), and an owner-scoped delete_message that mirrors existing repository conventions. The fairness requirement is met precisely: check_and_record runs before any mutation, so a rate-limited regenerate returns 429 and leaves the old answer intact while still consuming one audit row on success. The frontend mirrors this with optimistic removal and restore-on-failure, reusing the SSE parser and RateLimitError handling. Test coverage is thorough and meaningful across both tiers.

Weaknesses are minor and mostly cosmetic/process: the Regenerate button hardcodes inline hex colors and hover handlers rather than reusing the existing citation-button styling the plan called for, and — more notably — the PR body omits the prominent protected-path / human-review warning that the plan flagged as mandatory (messages.py and repository.py user_id-scoping are on the protected list). The code itself respects every security invariant, so this is a documentation/process gap rather than a correctness one. Overall an excellent cell result.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully delivers the issue: a regenerate action on the latest assistant message that deletes the stale answer, re-streams a fresh one in place over the identical SSE contract with its own citations, and runs check_and_record so the regenerate counts against the 25/24h cap. All four 'done looks like' criteria are met. |
| root_cause | 9/10 | Structurally sound: rather than duplicating ~130 lines, it extracts a shared _run_assistant_stream helper that both create_message and regenerate_message call, so the new path reuses the exact tool plumbing, refusal/citation logic, and shielded persistence. The ownership-scoped delete_message mirrors existing repository patterns. No band-aids. |
| scope_discipline | 9/10 | Surgical — touches exactly the 8 files the plan named and nothing else, with no drive-by refactors. Minor ding: the plan demanded the protected-path / human-review flag be surfaced prominently in the PR body, but the body omits it entirely. |
| subtle_correctness | 9/10 | Careful about failure modes: rate-limit check runs before any deletion so a 429 preserves the old answer; delete_message is owner-scoped via an EXISTS join; latest-assistant and preceding-user-turn guards return 400; the frontend optimistically removes the bubble and restores it on error/abort/429. Ordering and async/await are correct throughout. |
| code_quality | 8/10 | Clean, well-commented, idiomatic reuse of existing patterns and the runStream(url, body) factoring is tidy. Slight inconsistency: the Regenerate button uses ad-hoc inline hex colors (#94a3b8) and hand-rolled hover handlers instead of matching the existing citation-toggle button styling/CSS vars as the plan requested. |
| test_discipline | 9/10 | Strong, targeted coverage: backend happy path, usage-counted, 429-preserves-old-answer, ownership 404, and guard rails (non-latest, user message, no preceding user, empty conv); frontend hook tests for URL/no-body, sources parsing, and 429->RateLimitError, plus Message component render/click tests. Tests use real fixtures, not gamed assertions. |
| plan_impl_fidelity | 9/10 | Near line-for-line execution of the plan: identical delete_message SQL, the _run_assistant_stream(*, conv_id, user_id, current_user, llm_messages, first_user_message) signature, the exact endpoint guard ordering, the runStream extraction, and the full test matrix all match. The only divergence is the unflagged protected-path callout in the PR body. |

---

Issue #294, opened 2026-05-31
Filter and search conversation history by date and video


coleam00:
As a long-time user with a lot of past conversations, the sidebar search only matches conversation titles, which isn't enough to find things.

I want to find past conversations by **when** they happened (for example: this week, last month, or a custom date range) and by **which video** they were about, so I can get back to a discussion I only half-remember. I'd like these filters to combine with the text search I already have, and the results should still be ordered most-recent-first.

Done looks like: I can narrow my conversation list by a date range and by a video (optionally together with a text search), and see the matching conversations newest-first.

PR #304, opened 2026-05-31

coleam00:
Fixes #294

**Source run-id:** `5170e031-5f8f-418a-84a1-743101123f79`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -503,26 +503,63 @@ async def delete_conversation(conv_id: str, user_id: str) -> bool:
         return result != "DELETE 0"  # type: ignore[no-any-return]
 
 
-async def search_conversations_by_title(user_id: str, query: str, limit: int = 20) -> list[dict]:
-    """Return conversations owned by user where title contains substring (case-insensitive).
+async def search_conversations(
+    user_id: str,
+    query: str | None = None,
+    date_from: datetime | None = None,
+    date_to: datetime | None = None,
+    video_id: str | None = None,
+    limit: int = 20,
+) -> list[dict]:
+    """Return conversations owned by user, filtered by title, date range, and video.
 
-    Ported from the SQLite era: uses ILIKE in Postgres so the match is
-    case-insensitive in one step (no need to lower() both sides).
+    All filters are optional and compose with AND logic. Results ordered by
+    updated_at DESC.
     """
-    pattern = f"%{query}%"
-    async with _acquire() as conn:
-        rows = await conn.fetch(
-            """
-            SELECT id, title, created_at, updated_at
-            FROM conversations
-            WHERE user_id = $1 AND title ILIKE $2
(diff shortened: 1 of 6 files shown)


coleam00:
**Benchmark Evaluator** — score: **57/70**

A clean, well-scoped implementation that solves issue #294 end-to-end and tracks its plan almost exactly. The server-side query is the right architectural choice — a single composable WHERE clause with parameterized title/date/video filters and an EXISTS containment check against messages.sources — and the frontend wiring is careful (end-of-day dateTo normalization, stale-fetch guard, deps array, adding the preview column so the empty-conversation filter keeps working). Scope discipline is exemplary: only the files that needed changing were touched.

The one real weakness is test discipline. The repository test module remains entirely skipped (a pre-existing limitation pending a real test-Postgres, honestly documented), and the newly added date/video tests are aspirational placeholders — they assert outcomes that depend on back-dated created_at and seeded sources, but the seeding is left as comments, so the assertions would fail as written. So the new behavior ships with no executable verification. Everything else is production-grade.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers exactly what the issue asked: date-range + per-video filters that combine with the existing text search, results ordered updated_at DESC (newest-first). Only minor gap is no quick-pick presets (this week/last month), which the plan explicitly deferred. |
| root_cause | 9/10 | Structurally correct fix: replaces ad-hoc client-side title filtering with a single composable server-side query (conditional WHERE, EXISTS-on-sources JSONB containment). Not a band-aid — the filtering happens where the data lives and scales correctly. |
| scope_discipline | 10/10 | Touched precisely the five files the work required plus the test file — repository, route, api client, hook, Sidebar. Zero unrelated churn. |
| subtle_correctness | 8/10 | Thoughtful about edge cases: dateTo normalized to end-of-day UTC for inclusivity, the fetchIdRef race-guard preserved, useCallback deps updated, preview column added so the preview!==null filter stays consistent, all SQL parameterized. Minor: filters key off created_at while ordering is updated_at, a tolerable inconsistency. |
| code_quality | 8/10 | Readable dynamic WHERE builder with running param index, clean hook signature, idiomatic to the codebase. The heavy inline styles in Sidebar match the file's existing convention rather than introducing new patterns. |
| test_discipline | 4/10 | Ported title tests are real but the whole module stays skipped (pre-existing Postgres constraint, honestly noted). The new date/video tests are non-functional stubs — they create conversations but never seed the created_at/sources values their assertions depend on, so several would fail if un-skipped. No executable coverage of the new behavior. |
| plan_impl_fidelity | 9/10 | Implementation faithfully executes the plan across all five files: rename to search_conversations, conditional WHERE with $N placeholders, EXISTS JSONB filter, three optional route params, URLSearchParams client, hook signature change with end-of-day dateTo normalization, and a clear-filters button. Only the test work fell short of the plan's intent. |

---

PR #305, opened 2026-05-31

coleam00:
Fixes #294

**Source run-id:** `b0cba60e-e25f-4fc8-9a67-b7fc3152f85c`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -526,6 +526,90 @@ async def search_conversations_by_title(user_id: str, query: str, limit: int = 2
     return [dict(r) for r in rows]
 
 
+async def search_conversations(
+    user_id: str,
+    q: str | None = None,
+    start_date: str | None = None,
+    end_date: str | None = None,
+    video_id: str | None = None,
+) -> list[dict]:
+    """Filter a user's conversations by optional title text, date range, and video.
+
+    Every filter is optional and the non-null ones combine with AND. The result
+    mirrors ``list_conversations`` (same columns + ``preview`` subquery) and is
+    ordered most-recent-first. Filters are scoped to the owning user; a NULL
+    parameter disables that filter via the ``$N IS NULL OR ...`` guard so a
+    single prepared statement covers every combination.
+
+    - ``q``: case-insensitive substring match on the title (ILIKE).
+    - ``start_date`` / ``end_date``: ISO 8601 bounds on ``updated_at`` (inclusive),
+      passed straight through to Postgres TIMESTAMPTZ parsing.
+    - ``video_id``: keeps only conversations whose messages cite the given video,
+      extracted from the ``messages.sources`` JSONB citation array.
+    """
+    pattern = f"%{q}%" if q is not None else None
+    async with _acquire() as conn:
+        rows = await conn.fetch(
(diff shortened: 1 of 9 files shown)


coleam00:

A strong, well-targeted implementation that fully solves issue #294. It moves conversation filtering server-side with a single, cleanly-guarded SQL statement that combines title, date-range, and video filters while preserving the newest-first ordering and the response shape (preview) the frontend depends on. The frontend changes are idiomatic — a filter-object API wrapper, a hook that refetches correctly on value changes with a stale-response guard, and a Sidebar with sensible date presets and a video dropdown. Tests are honest: existing client-side-filter tests were rewritten to the new server-side contract rather than gamed, and new api/hook tests are meaningful.

Weaknesses are minor but real. The plan's LIMIT 50 bound was dropped, so the combined query can return an unbounded payload for users with very large histories. The backend tests, while well-constructed, are skipped pending a Postgres test harness, leaving the most failure-prone code (the JSONB video match) unverified in CI. The Sidebar's heavy inline styling deviates from the project's Tailwind-only convention, though it matches the file's existing style. None of these block the feature; collectively they keep an otherwise excellent PR just short of top marks.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers every part of the ask: date-range filter (this-week / last-month / custom), video filter, combined with text search via AND, all newest-first. The sidebar UI exposes all three and they compose exactly as the issue describes. |
| root_cause | 9/10 | Properly moves filtering server-side into a single parameterized SQL query with NULL-guarded optional filters, instead of band-aiding the existing client-side title match. JSONB sources are unnested correctly to back the video filter — a structurally sound design, not a workaround. |
| scope_discipline | 9/10 | Changes are confined to the relevant repository fn, route, api wrapper, hook, Sidebar, and their tests. The now-orphaned search_conversations_by_title is left untouched rather than opportunistically removed, which is the correct scope-disciplined call. |
| subtle_correctness | 8/10 | Thoughtful: stale-fetch guard preserved, effect deps destructured so it re-runs on value (not identity) changes, custom end-date pinned to end-of-day for inclusivity, falsy params omitted. One miss — the plan's LIMIT bound was dropped, leaving the query unbounded for heavy users. |
| code_quality | 8/10 | Clear naming, strong docstrings explaining the NULL-guard and JSONB walk, idiomatic asyncpg and React. The large inline-style blocks in Sidebar technically violate CLAUDE.md's Tailwind-only rule, but they match the existing file's prevailing style so they read consistently. |
| test_discipline | 7/10 | Frontend tests are real and targeted (api param encoding, hook server-side contract, refetch-on-change) and the existing useConversations tests were honestly rewritten to the new contract rather than gamed. Backend tests are well-written but skipped pending a Postgres harness, so the core SQL is not actually exercised in CI — consistent with repo convention but leaves the riskiest code unverified. |
| plan_impl_fidelity | 8/10 | Implementation faithfully executes the plan: both repo functions, route ordering before /{conv_id}, the filter-object api, server-side hook, and Sidebar controls all match. Two minor drifts — the planned LIMIT 50 bound was omitted, and the video match uses jsonb_array_elements + ->> rather than the suggested @> containment (functionally equivalent). |

---

PR #307, opened 2026-05-31
[bench-OO] review: fix ruff format on test_list_conversations_filters.py

coleam00:
Fixes #294

**Source run-id:** `6370bf0d-85d1-4672-a935-ceeb619c0ce8`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -450,22 +450,59 @@ async def get_conversation(conv_id: str, user_id: str) -> dict | None:
     return dict(row) if row else None
 
 
-async def list_conversations(user_id: str) -> list[dict]:
-    async with _acquire() as conn:
-        rows = await conn.fetch(
-            """
-            SELECT c.*,
-                   (SELECT content
-                    FROM messages
-                    WHERE conversation_id = c.id
-                    ORDER BY created_at DESC
-                    LIMIT 1) AS preview
-            FROM conversations c
-            WHERE c.user_id = $1
-            ORDER BY c.updated_at DESC
-            """,
-            user_id,
+async def list_conversations(
+    user_id: str,
+    *,
+    start_date: str | None = None,
+    end_date: str | None = None,
+    video_id: str | None = None,
+) -> list[dict]:
+    """List a user's conversations, newest-first.
+
+    Optional filters (combine with AND, all layered on top of the mandatory
+    owner scope):
+      - start_date / end_date: inclusive bounds compared against
+        ``updated_at`` (the "last activity" column the list is ordered by),
+        passed as ISO 8601 strings parsed by Postgres.
+      - video_id: only conversations that cite the given video in any
+        message's ``sources`` JSONB array, via ``@>`` containment.
+
+    Only placeholder *indices* are interpolated into the SQL text; every
+    *value* is bound through ``params`` (no f-string SQL injection).
+    """
+    conditions = ["c.user_id = $1"]
+    params: list[object] = [user_id]
+
+    if start_date is not None:
+        params.append(start_date)
+        conditions.append(f"c.updated_at >= ${len(params)}")
+    if end_date is not None:
+        params.append(end_date)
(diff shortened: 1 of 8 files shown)


coleam00:

A strong, well-executed PR. It solves issue #294 cleanly with the architecturally correct approach: because video membership lives in messages.sources and isn't returned by the list endpoint, filtering must happen server-side, and the implementation does exactly that via dynamically-built AND clauses (updated_at range + a JSONB-containment EXISTS subquery) layered on top of the untouched owner scope and newest-first ordering. The code is injection-safe (indices interpolated, values bound), idiomatic to the codebase, well-documented, and scope-disciplined. Frontend wiring is careful — stable filtersKey to avoid spurious refetches, a stale-fetch guard, inclusive UTC range widening, and graceful video-list fallback. Plan-to-implementation fidelity is essentially perfect.

Weaknesses are modest. The behavioral backend tests that would actually prove range/intersection/owner-scope correctness against a real Postgres are skipped stubs (consistent with the repo's existing skipped suites, and honestly labeled), so the core SQL behavior is only verified at the query-construction level. The preset date math mixes local-time formatting with UTC bounds, a benign off-by-one edge for far-from-UTC users. And although the change is purely additive, it touches protected paths (conversations.py, repository.py) that the factory rules require to be human-authored — the plan flagged this, but it remains a real gating risk for the PR as submitted.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers exactly what #294 asked: server-side date-range and video filters that combine with the existing text search, with presets (This week/Last month), a video dropdown, and results still ordered newest-first. Only minor gap is that the text search remains a client-side title-substring filter, but that's a deliberate, reasonable preservation of existing highlighting behavior. |
| root_cause | 9/10 | Structurally correct: video membership lives in messages.sources (never returned by the list endpoint), so client-side filtering is impossible — pushing filters into the SQL via an EXISTS JSONB-containment subquery and updated_at range is the right layer. Filters are additive AND-clauses on top of the mandatory owner scope, not a workaround. |
| scope_discipline | 9/10 | Surgical — exactly the 5 source files plus 2 test files, no unrelated 'improvements'. The one caveat is that conversations.py and repository.py are on the protected-paths list; the change is unavoidable and purely additive, and the plan correctly flagged it for human authorization, but a strict validator would still gate the PR. |
| subtle_correctness | 8/10 | Thoughtful: NULL sources naturally don't match (@> NULL is NULL), inclusive UTC range widening (T00:00:00.000Z / T23:59:59.999Z), placeholder indices interpolated but all values bound (no injection), stable JSON.stringify filtersKey to avoid identity-churn refetches, and the fetchIdRef stale-guard. Minor wrinkle: presets compute YYYY-MM-DD in local time then treat them as UTC bounds, so a user far from UTC can see a slight off-by-one-day edge, but it's benign for a coarse filter. |
| code_quality | 8/10 | Readable and idiomatic — matches the file's existing inline-style/Tailwind idiom, keeps all SQL in repository.py and all fetches in api.ts, and the repository docstring clearly documents semantics and the injection-safety invariant. The Sidebar inline styles are verbose, but that's consistent with surrounding code, so correctly not 'fixed'. |
| test_discipline | 7/10 | The Tier-1 query-construction tests (capturing fake conn) genuinely lock in placeholder numbering, owner scope, ordering, and value-binding, and the frontend tests for query-string serialization and hook refetch/stable-key behavior are solid and meaningful. The Tier-2 behavioral DB tests are skipped stubs with trivial isinstance asserts — honestly labeled and consistent with the existing skipped suites, but it means range/intersection/owner-scope behavior isn't actually exercised end-to-end. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — signature, keyword-only params, incremental conditions list, JSONB EXISTS containment, the exact api.ts/useConversations/Sidebar threading, preset buttons, empty-state copy update, and the mandated fixture-tier (not snapshot/wall-clock) date tests. No drift. |

---

PR #308, opened 2026-05-31
[bench-KO] fix(channels): don't count skipped videos as new; surface failed syncs

### LINKED ISSUE #295: Channel sync miscounts skipped videos as new and hides failed runs

coleam00:
Fixes #295

**Source run-id:** `112bc13c-4922-4dcc-9a31-a8765e1ed16d`

### DIFF
--- a/app/backend/routes/channels.py
+++ b/app/backend/routes/channels.py
@@ -156,7 +156,6 @@ async def sync_channel(limit: int | None = None, force: bool = False) -> SyncRes
             existing = await repo.get_video_by_youtube_id(youtube_video_id)
             if existing is not None and not force:
                 logger.info("Video %s already ingested, skipping", youtube_video_id)
-                videos_new += 1
                 await repo.update_sync_video_status(
                     video_id=sync_video_record["id"],
                     status="ingested",
--- a/app/backend/tests/test_channel_sync.py
+++ b/app/backend/tests/test_channel_sync.py
@@ -145,7 +145,7 @@ def make_mock_transcript(text):
 async def test_sync_channel_idempotent_skips_existing_videos():
     """
     If a video is already in the DB (matched by youtube_video_id in URL),
-    it is skipped and counted as 'new' (already ingested = already counted).
+    it is skipped and NOT counted as 'new' (only genuinely ingested videos count).
     """
     # Pre-ingest a video so get_video_by_youtube_id returns it
     await repository.create_video(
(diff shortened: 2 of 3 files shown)


coleam00:

An exemplary minimal fix. The PR correctly diagnoses that both reported symptoms share a single root cause — the spurious videos_new += 1 on the already-ingested skip branch — and that the existing status formula (failed iff videos_error > 0 and videos_new == 0) becomes correct the moment the counter stops being inflated. The one-line production change is the right fix, and the author resisted any temptation to touch the status logic, the force path, or the repository layer. Scope is tight (3 files, ~118 additions dominated by the new test).

The regression test is the standout: it reproduces the precise masking scenario from the issue (skipped + genuinely-new-all-fail) and asserts videos_new=0, videos_error=1, and status='failed', mocking the repository and Supadata boundaries in the established codebase style rather than requiring a DB. It also smartly lives in a new file to avoid the module-level skip on test_channel_sync.py. The only minor knocks are cosmetic plan drift in mock-target naming and the fact that test_channel_sync.py remains globally skipped (an out-of-scope infra issue the PR correctly declined to fix). Strong cell result for KO.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly fixes both reported bugs at once: removing the erroneous videos_new += 1 on the skip branch means skipped videos no longer inflate the 'new videos' count, and because the status formula keys off videos_new == 0, an all-new-fail run now correctly surfaces as 'failed'. |
| root_cause | 10/10 | True root cause, not a band-aid. The status-determination formula was already logically correct; the only defect was the inflated counter feeding it. A single-line deletion restores correctness to both symptoms with no compensating hacks. |
| scope_discipline | 10/10 | Surgical: one production line removed, one stale assertion corrected, one focused regression test file added. No unrelated refactoring, no churn in repository.py, schema, or the Supadata client. |
| subtle_correctness | 9/10 | Edge cases hold up: all-skipped runs yield videos_new=0/videos_error=0 -> 'completed' (correct), and the force=True re-ingest path that legitimately increments videos_new at line 339 is untouched. Partial success (some new succeed + some fail) still reads 'completed', which matches intended semantics. |
| code_quality | 9/10 | Idiomatic to the codebase — patches the route-module import bindings, uses httpx ASGITransport + dependency_overrides, and the new test is well-commented and clearly named. The corrected docstring/comment in the existing test keeps the intent honest. |
| test_discipline | 9/10 | Adds a meaningful regression test that exercises the exact issue scenario (skipped + all-new-fail -> videos_new=0, videos_error=1, status='failed'). The existing assertion change is legitimate (it previously documented the buggy behavior), not gaming. Placed in a separate file to dodge the module-level skip mark — a sound choice. |
| plan_impl_fidelity | 9/10 | Implementation faithfully executes the plan: line removed, existing assertion updated to videos_new==1, new regression file added with the specified asserts. Minor drift only in mock targets (patched channels.supadata.get_channel_video_ids / channels.repo.get_video_by_youtube_id rather than the plan's _get_client / repository module), which is consistent with the explore's documented import-binding pattern. |

---

PR #309, opened 2026-05-31

### LINKED ISSUE #295: Channel sync miscounts skipped videos as new and hides failed runs

coleam00:
Fixes #295

**Source run-id:** `73ad5c5c-990a-4e1c-835c-1402432d4c5d`

### DIFF
--- a/app/backend/routes/channels.py
+++ b/app/backend/routes/channels.py
@@ -156,7 +156,6 @@ async def sync_channel(limit: int | None = None, force: bool = False) -> SyncRes
             existing = await repo.get_video_by_youtube_id(youtube_video_id)
             if existing is not None and not force:
                 logger.info("Video %s already ingested, skipping", youtube_video_id)
-                videos_new += 1
                 await repo.update_sync_video_status(
                     video_id=sync_video_record["id"],
                     status="ingested",
--- a/app/backend/tests/test_channel_sync.py
+++ b/app/backend/tests/test_channel_sync.py
@@ -145,44 +145,89 @@ def make_mock_transcript(text):
 async def test_sync_channel_idempotent_skips_existing_videos():
     """
     If a video is already in the DB (matched by youtube_video_id in URL),
-    it is skipped and counted as 'new' (already ingested = already counted).
+    it is skipped and must NOT be counted as 'new'.
     """
-    # Pre-ingest a video so get_video_by_youtube_id returns it
-    await repository.create_video(
-        title="Already ingested",
-        description="Already in DB",
-        url="https://www.youtube.com/watch?v=dQw4w9WgXcQ",
-        transcript="Already ingested transcript.",
-    )
+    # Simulate one existing video and one new video
+    async def fake_get(youtube_video_id):
+        if youtube_video_id == "dQw4w9WgXcQ":
+            return {"id": "existing-video-id"}
+        return None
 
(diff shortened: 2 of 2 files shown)


coleam00:

A genuinely strong, minimal fix. The implementer (and the plan) correctly identified that a single misplaced `videos_new += 1` in the idempotent-skip branch was the shared root cause of both reported symptoms: it inflated the 'new videos' count and, by keeping videos_new non-zero, defeated the existing `videos_error > 0 and videos_new == 0` status check so that all-fail runs masqueraded as 'completed'. Deleting that one line fixes both at once — no schema changes, no new heuristics, no scope creep. Plan-to-implementation fidelity is essentially perfect, and the existing test assertion was corrected honestly rather than gamed.

The one real weakness is test discipline: the regression test is well-constructed and would correctly fail before the fix and pass after, but the whole test_channel_sync.py module is skipped via `pytestmark = pytest.mark.skip` (broken temp_db_schema fixture pending an Alembic rewrite). As shipped, the new test never executes, so it provides no live protection against regression and this isn't surfaced in the PR body. The plan flagged lifting the skip as pre-existing out-of-scope debt, which is a reasonable call, but it leaves the issue's 'add a regression test' requirement only nominally met. Overall a clean, high-quality cell held back from top marks by an inert test.

## Scores

|---|---|---|
| solves_issue | 9/10 | Removing the stray `videos_new += 1` in the skip branch addresses both halves of the issue at once: the new-video count now reflects only truly-ingested videos, and the existing status check (`videos_error > 0 and videos_new == 0`) now correctly surfaces a run where all genuinely-new videos fail as 'failed'. Docked one point only because the accompanying regression test is inert (see test_discipline). |
| root_cause | 9/10 | Textbook root-cause fix — a single misplaced counter increment was the common origin of both symptoms (inflated count AND hidden failure via the poisoned status heuristic). No band-aid; the structural cause is removed and the downstream status logic naturally produces correct results. |
| scope_discipline | 10/10 | Perfectly surgical: one line deleted in channels.py plus targeted test changes. No unrelated files touched, no opportunistic refactors, force-path and success-path left intact. |
| subtle_correctness | 9/10 | Correctly reasons about the interaction between the counter and the status heuristic; all-skip runs still resolve to 'completed' (videos_error==0) and partial failures remain 'completed', which matches the issue's intent. No async/type foot-guns introduced. Minor latent fragility in the 'failed' heuristic itself (partial failures stay 'completed') is pre-existing and out of scope. |
| code_quality | 8/10 | The source change is a clean deletion. Test code follows existing file conventions, though the deeply-nested with-blocks are a bit verbose. Production-grade for the change at hand. |
| test_discipline | 5/10 | A well-targeted regression test was added that genuinely distinguishes fixed vs. unfixed behavior, and the existing assertion was honestly corrected (not gamed) to videos_new==1. BUT the entire test file sits under a module-level `pytestmark = pytest.mark.skip`, so neither test actually runs in CI — the regression is not truly guarded. The PR body doesn't flag this. The plan declared lifting the skip out of scope, which is defensible, but shipping a skipped regression test only weakly satisfies 'add a regression test.' |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim: remove the single increment, add `test_sync_channel_skips_do_not_count_as_new_and_failed_run_is_surfaced` with the exact asserted values (total=3, new=0, error=1, status='failed'), and update the existing idempotency test from videos_new==2 to ==1. No drift. |

---

PR #310, opened 2026-05-31

### LINKED ISSUE #295: Channel sync miscounts skipped videos as new and hides failed runs

coleam00:
Fixes #295

**Source run-id:** `374a3ef5-4189-4ca2-8b1e-c7ae484df14b`

### DIFF
--- a/app/backend/alembic/versions/0001_initial.py
+++ b/app/backend/alembic/versions/0001_initial.py
@@ -159,7 +159,7 @@ def upgrade() -> None:
             id TEXT PRIMARY KEY,
             sync_run_id TEXT NOT NULL REFERENCES channel_sync_runs(id) ON DELETE CASCADE,
             youtube_video_id TEXT NOT NULL,
-            status TEXT NOT NULL CHECK (status IN ('pending', 'ingested', 'error')),
+            status TEXT NOT NULL CHECK (status IN ('pending', 'ingested', 'error', 'skipped')),
             error_message TEXT,
             created_at TIMESTAMPTZ NOT NULL
         )
--- a/app/backend/alembic/versions/0006_add_skipped_sync_status.py
+++ b/app/backend/alembic/versions/0006_add_skipped_sync_status.py
@@ -1,0 +1,39 @@
+"""Add 'skipped' to channel_sync_videos status CHECK constraint.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    # PostgreSQL auto-generates the constraint name as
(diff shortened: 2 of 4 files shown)


coleam00:
**Benchmark Evaluator** — score: **54/70**

A technically strong implementation let down by inert tests. The source fix is exactly right: it removes the root-cause miscount in the skip branch (videos_new no longer padded by already-ingested videos), introduces a distinct 'skipped' status, and leaves the overall-status formula untouched so it now correctly reports 'failed' when every genuinely-new video errors. Notably, the implementer (Kimi) caught a footgun the Opus plan missed — writing the new 'skipped' value would violate the existing CHECK constraint — and correctly added a well-formed Alembic migration (0006) plus a matching update to 0001_initial, a justified override of the plan's overly-strict 'no schema changes' instruction. The migration chain, idempotency, and downgrade are all sound, and no downstream consumer filters on the literal 'ingested' status.

The glaring weakness is test discipline. Both the corrected existing assertion and the new regression test live under a module-level pytest.mark.skip that the implementer did not lift, so neither test actually runs — the regression test the issue specifically requested is dead code. The plan called out this exact risk and instructed the implementer to either remove the skip (if the test avoids the broken fixture) or document the limitation in the PR body; neither happened, and the PR body is boilerplate. The change is very likely correct, but it ships effectively unverified by the suite.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully addresses both symptoms: skips now increment videos_skipped (not videos_new), so the count reflects only truly-new ingests, and the untouched status formula correctly surfaces 'failed' when all genuinely-new videos error. The added migration also ensures status='skipped' doesn't blow up at runtime. |
| root_cause | 9/10 | Fixes the single source miscount in the skip branch rather than patching the downstream status formula — exactly the right structural fix. The status formula stays correct for all-skipped, all-failed, and mixed cases without modification. |
| scope_discipline | 9/10 | Surgical: channels.py source change, the relevant test file, and two migration files. The migrations exceed the plan's stated 'out of scope' but were strictly necessary for the new 'skipped' value to be writable; no unrelated cleanups crept in. |
| subtle_correctness | 9/10 | Caught the CHECK-constraint footgun the plan missed — without migration 0006 the update_sync_video_status(status='skipped') call would raise an IntegrityError. Migration chain (0005->0006), idempotent DROP IF EXISTS, and a real downgrade are all correct. |
| code_quality | 9/10 | Clean and idiomatic: consistently named videos_skipped counter, updated completion log line, well-documented migration matching the repo's Alembic style. Reads like the surrounding code. |
| test_discipline | 3/10 | The test content is meaningful and well-targeted, but the module-level pytest.mark.skip (line 44) is still present, so both the corrected assertion and the new regression test are SKIPPED and never execute. The plan explicitly warned to lift the skip or note it in the PR body; neither was done, so the regression test the issue demanded is inert and unverified. |
| plan_impl_fidelity | 6/10 | Core change faithfully executes the plan (drop videos_new++, set status='skipped', add counter, update log, leave formula untouched). One justified positive deviation (necessary migration the plan wrongly excluded) and one real miss: the impl ignored the plan's explicit test-skip caveat, leaving the new test non-running and undocumented. |

---

PR #311, opened 2026-05-31
[bench-OO] fix(channels): stop counting skipped videos as new in sync

### LINKED ISSUE #295: Channel sync miscounts skipped videos as new and hides failed runs

coleam00:
Fixes #295

**Source run-id:** `dee56a15-680e-4e4f-80c8-057772c30d26`

### DIFF
--- a/app/backend/routes/channels.py
+++ b/app/backend/routes/channels.py
@@ -156,7 +156,6 @@ async def sync_channel(limit: int | None = None, force: bool = False) -> SyncRes
             existing = await repo.get_video_by_youtube_id(youtube_video_id)
             if existing is not None and not force:
                 logger.info("Video %s already ingested, skipping", youtube_video_id)
-                videos_new += 1
                 await repo.update_sync_video_status(
                     video_id=sync_video_record["id"],
                     status="ingested",
--- a/app/backend/tests/test_channel_sync_counts.py
+++ b/app/backend/tests/test_channel_sync_counts.py
@@ -1,0 +1,164 @@
+"""
+Regression tests for issue #295 — channel sync miscounts skipped videos as new
+and hides failed runs.
+
+Two bugs, one root cause: the skip branch in `sync_channel` used to increment
+`videos_new` for videos that already existed and were *not* re-ingested. That
+(1) inflated the "new videos" figure on routine no-op syncs and (2) masked real
+ingestion outages, because the end-of-run status check
+(`"failed" if videos_error > 0 and videos_new == 0`) never saw `videos_new == 0`
+when some videos had been skipped.
+
+These tests mock the `repo` layer (and Supadata / the ingest helper) so they run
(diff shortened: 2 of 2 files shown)


coleam00:

This PR is a textbook example of a correct, disciplined fix. It identifies that two distinct reported symptoms (inflated 'new videos' count on no-op syncs, and failed runs masquerading as 'completed') stem from a single root cause — the skip branch incrementing videos_new for videos that were never re-ingested — and removes exactly the one offending line. Because the end-of-run status logic was already correct, no other production change is needed, and the fix correctly preserves the per-video 'ingested' record and the force=True re-sync path. Scope is impeccably surgical: one deletion plus a new test file, with the out-of-scope items (skipped SQLite test module, schema/counter additions) deliberately untouched.

The regression tests are the standout: both meaningfully cover the fixed behavior, would have failed before the change, and assert against both the HTTP response and the persisted sync-run record while running without a live database via clean repo-layer mocking. The implementation faithfully mirrors the plan down to the test names and mock strategy. The only minor nitpicks are some duplicated mock-setup boilerplate across the two tests; these are cosmetic. Overall a high-quality, production-grade change.

## Scores

|---|---|---|
| solves_issue | 10/10 | Removing the single `videos_new += 1` in the skip branch fixes both reported symptoms at once: routine no-op syncs no longer inflate the new-video count, and runs where all genuinely-new videos fail now have videos_new==0 so the existing status check correctly reports 'failed' instead of hiding the outage. |
| root_cause | 10/10 | This is the true root cause, not a band-aid: the erroneous increment was the shared source of both bugs, and once removed the status line `"failed" if videos_error > 0 and videos_new == 0` becomes correct automatically with no further change. |
| scope_discipline | 10/10 | One line deleted in production code plus one new dedicated test file — nothing else touched. The skipped test_channel_sync.py, schema, counters, and force-path were all correctly left alone per the out-of-scope notes. |
| subtle_correctness | 9/10 | Correctly preserves the per-video 'ingested' status for skips (they are in the library) while only changing the run-level counter, and leaves the force=True path (which legitimately increments later) untouched. No async/await or type foot-guns introduced. |
| code_quality | 9/10 | The fix is minimal and idiomatic; the new test module is well-documented, clearly explains the two-bugs-one-cause reasoning, and uses AsyncMock/patch cleanly matching the existing test_channel_sync.py style. Minor: some mock setup boilerplate is duplicated between the two tests rather than shared via a helper. |
| test_discipline | 10/10 | Two targeted regression tests that genuinely exercise the fix and would both fail pre-fix (test 1 would see videos_new==2, test 2 videos_new==1/status completed). They assert both the JSON response and the persisted update_sync_run kwargs, and run without Postgres by mocking the repo layer. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan exactly: the single-line deletion, the unchanged status line, the untouched force path, and the two named regression tests with their mock strategy all match the plan's specific changes and test descriptions precisely. |

---

PR #312, opened 2026-05-31
[bench-XK] review: update stale inline comment in _collapse_by_video call site

### LINKED ISSUE #276: Missing source citation when an answer uses two nearby moments from the same video

coleam00:
Fixes #276

**Source run-id:** `c9b3b11f-188a-431f-99fe-4f0ee377c871`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -467,22 +467,17 @@ async def _maybe_set_conversation_title(
 
 
 def _collapse_by_video(chunks: list[dict[str, Any]]) -> list[dict[str, Any]]:
-    """Collapse multiple chunks from the same video into a single citation entry.
-
-    After the ``is_cited`` pass, chunks from the same video are redundant in
-    the UI — the user needs one clickable chip per video, not one per 5-second
-    transcript segment (issue #208).
+    """Collapse uncited chunks from the same video into a single citation entry,
+    but keep each cited chunk as its own entry so distinct moments referenced
+    by the model each get their own source chip (issues #208, #276).
 
     For each ``video_id`` group:
-    - ``is_cited`` is True if ANY chunk in the group was cited by the LLM.
-    - ``start_seconds`` is the earliest timestamp among cited chunks (or among
-      all chunks if none were cited), so the deep-link opens near the most
-      relevant moment.
-    - ``segment_count`` records how many chunks were collapsed so the frontend
-      can optionally display "(N segments)".
-    - All other fields are taken from the representative chunk.
-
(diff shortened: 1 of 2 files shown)


coleam00:

A clean, well-targeted fix that nails the root cause. The original _collapse_by_video collapsed every same-video chunk into a single earliest-timestamp representative, which simultaneously explains both reported symptoms (a missing second chip and a surviving chip pointing at the earlier moment). The rewrite splits the group into one entry per cited chunk plus one collapsed entry for all uncited chunks, preserving the prior #208 (de-noise uncited segments) and #176 (cap fallback) behaviors. Scope is tight, the per-chunk start_seconds resolves the wrong-timestamp case, and a focused regression test plus sensible updates to the two adjacent tests give good coverage. Weaknesses are minor: persistence/reload parity is reasoned about in the exploration rather than asserted by a dedicated test, and plan-fidelity is unscorable because this no-plan cell produced no planning artifact. Overall a strong PR for an impl-only (no-plan) baseline.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes the reported bug: each cited chunk from the same video now produces its own chip carrying its own start_seconds, so two nearby cited moments both surface with correct timestamps. Since persisted sources are the post-collapse structure, both the live-stream and reload paths are fixed. |
| root_cause | 9/10 | Addresses the true root cause — _collapse_by_video unconditionally collapsed an entire video group to one min-timestamp representative, discarding other cited chunks. The rewrite keeps cited chunks distinct while still collapsing uncited noise (preserving #208) and the cap fallback (#176). |
| scope_discipline | 10/10 | Surgical: only the offending function plus its stale call-site comment in messages.py and the directly-affected tests. No unrelated churn (80/-47 across 2 files). |
| subtle_correctness | 9/10 | Per-chunk start_seconds avoids the wrong-timestamp symptom; insertion order and cited-before-uncited ordering preserved; segment_count semantics handled sensibly (1 per cited, count-of-absorbed for uncited); None timestamps guarded with `or 0.0`. Cap fallback still lets cited entries pass through unbounded as intended. |
| code_quality | 9/10 | Readable, idiomatic to the module, clear list-comprehension split of cited/uncited, and the docstring was updated to match the new contract. Matches surrounding style. |
| test_discipline | 9/10 | Existing tests were updated to reflect the corrected semantics (not gamed), and a dedicated #276 regression test asserts two cited moments from one video yield two correctly-timestamped chips plus one collapsed uncited entry. Targeted and meaningful. |
| plan_impl_fidelity | 5/10 | Not visible (source-plan.md not present) — this is the plan=NONE (XK) cell, so no planning artifact exists to compare against. |

---

Issue #147, opened 2026-04-23
Feature: gated Dynamous content + Circle membership verification (Apr 27)


coleam00:
## Context

Add a major new feature to DynaChat (the app running at `chat.dynamous.ai` today): **gated retrieval over Dynamous courses and workshops, verified against paid Circle membership**. The existing YouTube RAG keeps working exactly as it does today — we're layering a second content source on top and adding a per-user access gate. No new site, no rewrite, no rebrand.

Ships as part of the Dynamous 1-year anniversary on **2026-04-27** (hard deadline, 4 days from filing).

This issue is `factory:needs-human` on purpose — it's a comprehensive plan, not an autonomous work unit. Work will be done manually by Cole in a future session. Do not auto-triage or auto-implement.

## What stays exactly as it is today

- `chat.dynamous.ai` remains the only URL
- DynaChat name, branding, and UI shell
- Existing users, conversations, messages, rate limit (25/24h)
- Email+password signup + login, JWT session cookies
- YouTube ingestion pipeline (Supadata), MiniLM embeddings, pgvector storage, RRF hybrid retrieval, blue/green deploy

## What this feature adds

1. A second content type alongside YouTube: Dynamous course and workshop transcripts
2. A per-user `is_member` flag, verified against Circle's Admin API
3. Retrieval that filters by `source_type` based on `is_member`: paid members see YouTube + Dynamous; everyone else sees YouTube only
4. A content pipeline for transcribing and ingesting the Dynamous videos

## Behavior matrix

| User | What retrieval returns |
|---|---|
| Signed-in paid Dynamous member | YouTube + Dynamous course/workshop chunks |
| Signed-in, not a paid member (never was, or cancelled) | YouTube chunks only (silent — no upsell banner) |
| Not signed in | Login page — no access, same as today |

Membership is verified: on signup, on login, and on `/me` requests where `member_verified_at` is older than 1 hour. No webhooks, no cron.

## Decisions locked (from 2026-04-23 planning session)

| Topic | Decision |
|---|---|
| Deadline | 2026-04-27 hard |
| Launch scope | All courses (Second Brain, AI Agent Mastery, Agentic Coding) + all workshops |
| Deploy | Same deployment, same URL — `chat.dynamous.ai`. Standard blue/green. |
| Branding | No rebrand. Keep DynaChat. |
| Repo layout | This repo (dark-factory-experiment) stays public. Transcripts live in a new **private** repo `coleam00/dynamous-content`. |
[...]

PR #313, opened 2026-05-31
[bench-XO] fix(citations): keep distinct same-video moments as separate chips

### LINKED ISSUE #276: Missing source citation when an answer uses two nearby moments from the same video

coleam00:
Fixes #276

**Source run-id:** `d150078a-adb7-4adf-b1ad-5c6ebc59b711`

### DIFF
--- a/app/backend/rag/expansion.py
+++ b/app/backend/rag/expansion.py
@@ -34,13 +34,13 @@ async def expand_and_merge(
                 Defaults to repository.get_chunk_neighbors.
 
     Returns:
-        List of span dicts (same shape as input chunks but with merged content):
-          - video_id, video_title, video_url from the video of the first chunk in span
-          - content: concatenated text of all chunks in the span
-          - start_seconds: from the first chunk in the span
-          - end_seconds: from the last chunk in the span
-          - snippet: from the originally-retrieved chunk (preserved for citation)
-          - chunk_id: from the originally-retrieved chunk (preserved for citation)
+        List of citation dicts (same shape as input chunks but with merged
+        content). A contiguous span emits ONE entry per originally-retrieved
+        chunk it contains (issue #276), each:
+          - video_id, video_title, video_url from the span's first chunk
+          - content: concatenated text of all chunks in the span (shared context)
+          - start_seconds / end_seconds: from this entry's own retrieved chunk
+          - snippet / chunk_id: from this entry's own retrieved chunk
     """
     if window <= 0 or not chunks:
         return chunks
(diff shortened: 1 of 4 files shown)


coleam00:

A strong, well-reasoned fix for a no-plan (impl=Opus) cell. Its standout quality is root-cause depth: the implementation independently identified that the bug originates in TWO places — expand_and_merge collapsing nearby retrieved chunks into a single span anchored to only the first chunk, and _collapse_by_video grouping by video_id — even though the exploration artifact flagged only the latter. Fixing the expansion layer is precisely what makes the bug 'nearby-moment' specific, so this is the correct comprehensive fix rather than a symptom patch. The change is scoped tightly to the citation path, preserves the SSE/Citation contract (no frontend or persistence migration needed), respects prior invariants (issue #147 source_type/lesson_url, tuple keying), and ships meaningful regression tests at both layers with honestly-updated existing assertions. Weaknesses are minor: sibling chips from one expanded span carry duplicated merged content text, and segment_count semantics quietly shift (cited chips are always 1 now) without a frontend note. Plan-fidelity is unscorable since this is the no-plan baseline.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes the reported behavior: distinct same-video moments now each keep their own chip with their own correct timestamp, and a regression test is added as requested. Addresses both the 'missing chip' and 'wrong (earlier) timestamp' symptoms. |
| root_cause | 9/10 | Finds the true two-layer root cause: expand_and_merge collapsed two nearby retrieved chunks (overlapping neighbor windows) into one span anchored to the first chunk, and _collapse_by_video then grouped purely by video_id. Notably the impl uncovered the expansion-layer cause that the exploration doc never identified — that layer is what makes the bug specific to 'nearby' moments. |
| scope_discipline | 9/10 | Surgical: only the two source files in the citation path (expansion.py, messages.py) plus their two test files. No frontend churn — the Citation shape is preserved, so no renderer changes were needed. No unrelated 'improvements'. |
| subtle_correctness | 8/10 | Careful about subtle modes: per-anchor timestamp/snippet/chunk_id, defensive empty-anchors fallback, preserves issue #147 source_type/lesson_url from the hydrated anchor, and keeps tuple (video_id,chunk_index) keying. Minor wart: sibling chips from one span share identical merged content text, but it's harmless context, not a bug. |
| code_quality | 9/10 | Comments explain the why (issue refs, anchor rationale, drop-uncited-siblings reasoning), naming is clear (anchors, cited_video_ids, uncited_groups), and the rewrite of _collapse_by_video reads cleanly and idiomatically for this codebase. |
| test_discipline | 9/10 | Two new targeted tests reproduce the exact bug at both layers (two cited moments same video each get own chip; two nearby retrieved chunks in one contiguous span keep distinct anchors). Existing tests were updated to assert the corrected semantics (own-timestamp, segment_count=1 for cited) rather than gamed to pass. |
| plan_impl_fidelity | 5/10 | Not visible (source-plan.md not present) — this is the plan=NONE baseline cell, so there is no plan to compare against. Scored 5 per instructions. |

---

PR #314, opened 2026-05-31
[bench-XK] impl: address issue #277 — saved answer's sources don't match what was shown live after an interrupted response

coleam00:
Fixes #277

**Source run-id:** `b310f52c-6eef-4f42-b9aa-b5e690854f61`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -174,6 +174,12 @@ async def create_message(
         # Two-tier citations (issue #176): strip `[c:<id>]` markers from the
         # stream; use them at [DONE] to flag is_cited on retrieved chunks.
         marker_stripper = CitationMarkerStripper()
+        # Guard against persisting sources that were never successfully emitted
+        # to the client (issue #277).  The flag is only set after a yield
+        # returns normally; if GeneratorExit is raised at the yield point
+        # (client disconnect) the flag stays False.
+        sources_emitted = False
+        sources_snapshot: list[dict] | None = None
         try:
             # is_member_for_turn is captured above when tools are wired.
             # Re-bind to a local that's always defined so we can pass it
@@ -227,8 +233,10 @@ async def create_message(
                             else _extract_text_from_sse(full_response)
                         )
                         if not _is_refusal(final_text):
+                            sources_snapshot = list(source_citations)
                             sources_json = json.dumps(source_citations)
                             yield f"event: sources\ndata: {sources_json}\n\n"
+                            sources_emitted = True
                     full_response.append(sse_chunk)
                     yield sse_chunk
                     continue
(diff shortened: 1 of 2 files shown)


coleam00:

An excellent, well-scoped fix that correctly identifies and resolves the actual root cause of issue #277. Tracing the pre-fix code, the [DONE] handler populates source_citations, collapses same-video chunks in place via _collapse_by_video, then emits the SSE sources event — while the finally block independently re-reads source_citations to persist. If the stream is interrupted or _collapse_by_video raises mid-DONE, the list is left un-collapsed/partially mutated yet still persisted, producing exactly the longer, duplicate-per-video, sometimes-phantom chips on reload that disagree with the live view. The PR snapshots the list at the emit point (post-collapse, equal to the wire payload) and persists only that snapshot, gated on a sources_emitted flag that stays False on disconnect or exception. The shallow-copy is safe here because the list is not mutated after the snapshot, and the GeneratorExit handling is sound.

Scope is exemplary (two files, six logic lines) and the regression suite is strong, especially the test that forces _collapse_by_video to raise — directly reproducing the bug mechanism rather than a generic interrupt. Weaknesses are minor: the clean-path test uses distinct video_ids so it doesn't explicitly prove wire==persist on a collapsible set, and a rare benign race exists where a client could receive the sources event yet persistence falls back to None (the safe direction). Plan-fidelity is not scorable because source-plan.md is absent (no-plan XK baseline), scored 5 per instructions. A high-quality result for a no-plan cell.

## Scores

|---|---|---|
| solves_issue | 9/10 | Precisely fixes the reported scenario: it snapshots source_citations at the exact emit point and persists only that snapshot, gated on a sources_emitted flag, so an interrupted/errored mid-stream answer no longer saves a longer, duplicate-per-video, or phantom source list that disagrees with the live view. Normal completions remain unchanged. |
| root_cause | 9/10 | Correctly diagnoses the true root cause: the finally block re-reads source_citations, which inside the [DONE] handler is mutated in place by _collapse_by_video AFTER population — so an exception or interrupt mid-DONE leaves an un-collapsed list that gets persisted while the wire shows something else (or nothing). Snapshot-at-emit plus an emitted-flag is the structurally sound fix, not a band-aid. |
| scope_discipline | 10/10 | Surgical: only routes/messages.py (six lines of logic) and its dedicated test file. No unrelated churn or refactors. |
| subtle_correctness | 8/10 | The snapshot is taken at the right moment — after _collapse_by_video and the cap, identical to the json.dumps payload on the wire — and source_citations is not mutated again afterward, so the shallow list() copy is safe. GeneratorExit reasoning is sound (flag set only after the yield returns). Only a rare benign race remains: if GeneratorExit hits at the sources yield the client may have received the event yet persistence is None, which is the safe direction for this issue. |
| code_quality | 8/10 | Clear, idiomatic, well-named (sources_emitted / sources_snapshot) with comments that explain the why and cite issue #277. Matches the heavily-commented style of the surrounding handler. |
| test_discipline | 9/10 | Three meaningful regression tests: clean completion (persist == SSE), client disconnect before [DONE] (no persist), and — crucially — an exception inside _collapse_by_video during [DONE] processing, which reproduces the actual root-cause mechanism (partially-mutated list must not persist). Realistic fakes, no gamed tests. Minor: the clean test uses distinct video_ids so it doesn't directly assert dedup-equivalence on a collapsible set. |
| plan_impl_fidelity | 5/10 | Not visible (source-plan.md not present). The artifacts dir contains only source-exploration.md and .pr-number; no plan file exists to compare against. Consistent with the XK cell being a deliberate no-plan baseline. |

---

PR #315, opened 2026-06-01
[bench-XO] impl: persist sources identically to the live stream (#277)

coleam00:
Fixes #277

**Source run-id:** `32f101c6-e8df-41c9-8ba5-a81e2f030c0c`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -122,7 +122,6 @@ async def create_message(
     # lists exactly what the model actually read. The video_id whitelist is
     # only consulted by the transcript tool (it guards against hallucinated
     # ids); the search tools ignore it.
-    source_citations: list[dict] = []
     tool_chunks_acc: list[dict] = []
     embedding_cache: dict[str, list[float]] = {}
     tools_param: list[dict] | None = None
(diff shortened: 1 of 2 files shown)


coleam00:

A strong no-plan/Opus-impl PR. The fix correctly identifies that the live-vs-reload divergence stemmed from two independent derivations of the source list, and replaces them with one `_finalize_source_citations` helper whose result is both streamed and persisted. Crucially it captures the persist decision at the single emit point, before the yield that can raise CancelledError on disconnect, so the saved sources equal what the user saw whether the answer completed cleanly or was interrupted. Scope is surgical and the regression tests are well-aimed, covering the normal path, the errored-before-[DONE] path, and the cancel-at-yield window. Weaknesses are minor: the exact baseline mechanism that would produce the reported 'longer/duplicated chips on reload' is somewhat narrow (the exploration itself notes the plain error path was already consistent), so part of the win is hardening/maintainability rather than a previously-reproducible crash; and the tests lean on small sleeps for the shielded save. Plan fidelity is not scorable since this cell ran without a planning node.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly delivers the requested invariant: persisted sources are now a verbatim copy of the finalized set chosen at the single emit point, so reload can never show a longer/duplicated/refusal-attached chip set than the live stream. The interrupted-before-[DONE] case correctly persists None. |
| root_cause | 9/10 | Genuine structural fix, not a band-aid: it extracts a single `_finalize_source_citations` source of truth shared by both the SSE and persist paths and captures the persist snapshot BEFORE the cancellable `yield`, closing the disconnect-timing divergence window the exploration identified. The old dual-derivation design was the actual cause. |
| scope_discipline | 10/10 | Surgical: one production file (messages.py) refactored and one new test file added. No unrelated changes, no drive-by edits to footguns or other modules. |
| subtle_correctness | 9/10 | Thoughtful about the hard parts — CancelledError/asyncio.shield timing, capturing the decision before the yield that can raise, and keeping refusal suppression symmetric across both paths (an asymmetry the old fallback could introduce). No obvious foot-guns; types are correct and `Any` is already imported. |
| code_quality | 9/10 | Clean, well-named helper with a clear docstring enumerating the ordered finalization steps; idiomatic to the codebase and matches the existing `_collapse_by_video` style. Comments are slightly verbose but accurate and explain the non-obvious timing rationale. |
| test_discipline | 9/10 | Meaningful, targeted regression tests: a pure-helper unit test, a persisted==streamed deep-equality test, the interrupted-before-[DONE] -> None case, and a cancel-exactly-at-sources-yield snapshot test. They pin the exact invariant rather than gaming existing assertions; only minor reliance on sleep(0.05) to let the shielded save settle. |
| plan_impl_fidelity | 5/10 | Not visible (source-plan.md not present) — expected for the XO no-plan cell. Scored neutral per instructions; only the explore artifact exists and the impl is consistent with it. |

---

PR #317, opened 2026-06-01
[bench-XO] review: fix infinite re-fetch loop on error in VideoScopePicker

coleam00:
Fixes #279

**Source run-id:** `08173b7a-adcf-4ad4-a00c-56471afbd8d1`

### DIFF
--- a/app/backend/alembic/versions/0006_add_video_scope_to_conversations.py
+++ b/app/backend/alembic/versions/0006_add_video_scope_to_conversations.py
@@ -1,0 +1,40 @@
+"""Add per-conversation video scope (issue #279).
+
+Lets a user restrict a conversation to a subset of videos so the assistant's
+answers and citations only draw from those videos. The scope persists for the
+life of the conversation.
+
+`video_scope` is a nullable TEXT[] of video ids:
+  - NULL (default) → unscoped; retrieval searches the whole library (today's
+    behaviour, unchanged for every existing conversation).
+  - non-empty array → retrieval is restricted to those video ids.
+
+TEXT[] (not JSONB) so the scope reads back as a native Python list with no
+codec dance and slots straight into the existing `video_id = ANY($n::text[])`
+SQL filter pattern used by keyword/vector search.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    # Nullable, no default — NULL means "unscoped" so existing conversations
+    # keep searching the whole library.
+    op.execute("ALTER TABLE conversations ADD COLUMN IF NOT EXISTS video_scope TEXT[]")
+
+
+def downgrade() -> None:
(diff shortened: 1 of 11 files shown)


coleam00:
**Benchmark Evaluator** — score: **57/70**

A strong, well-engineered no-plan (Opus impl) PR. It solves issue #279 end-to-end: per-conversation video scope is stored on the conversation, persists across the conversation's life, and is enforced uniformly across hybrid/keyword/semantic search plus a dedicated transcript guard, while leaving the unscoped default path byte-for-byte unchanged. The work is structurally sound (filter pushed to the SQL layer using the codebase's existing ANY-filter idiom), carefully backward-compatible (kwargs gated so old test fakes keep working), and the self-review caught a real frontend bug (the useEffect re-fetch loop named in the PR title). Documentation and tests are above average.

Weaknesses are modest: it touches two protected files (routes/conversations.py, routes/messages.py) which a strict factory gate would auto-reject even though the edits are non-auth; there is no integration test for the update_conversation_scope SQL/owner-scoping; and the mid-conversation re-scope API exists but isn't wired into the UI beyond the landing picker. Plan fidelity is unscorable (no plan produced in this cell). Net: a high-quality implementation that would merge with only the protected-path caveat to resolve.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully delivers the ask: a landing-screen picker scopes a new conversation to chosen videos, the scope is persisted on the conversation row and re-read every turn (so it sticks for the conversation's life), and every retrieval tool plus the transcript tool is restricted to those ids. Empty selection falls through to whole-library search exactly as before. A scope indicator surfaces the restriction in-chat. |
| root_cause | 9/10 | Structurally sound — the scope is threaded at the correct layer (SQL WHERE clause via the existing `= ANY($n)` pattern) from migration -> repository -> hybrid retriever -> tool executors -> route, not faked at one chokepoint. The transcript tool gets its own guard so the model cannot bypass scoped search by requesting a full out-of-scope transcript. |
| scope_discipline | 8/10 | Every change is directly tied to the feature; no unrelated churn. Minor deduction: it edits routes/conversations.py and routes/messages.py, both on CLAUDE.md's protected-paths list (must be human-authored). The edits are non-auth and the exploration flagged this, but a strict factory gate would still trip on it. |
| subtle_correctness | 9/10 | Thoughtful about failure modes: empty list normalized to NULL consistently in DB/route/frontend, kwargs forwarded only when scope is set to preserve the pre-#279 signature for existing test fakes, stale/deleted ids degrade gracefully (match nothing), SELECT * propagates the new column, and the useEffect infinite re-fetch loop was caught and fixed in self-review. f-string SQL only interpolates a placeholder index, so it stays injection-safe. |
| code_quality | 9/10 | Production-grade and idiomatic to this codebase — docstrings explain the why, comments justify the non-obvious (deps omission, kwarg gating), naming matches existing conventions, and it reuses the established allowed_source_types filtering pattern rather than inventing a new one. |
| test_discipline | 8/10 | Solid targeted tests: normalize-scope helper (trim/dedupe/empty), structural signature guards, scope forwarding to both retrieval legs and the dispatcher, transcript in/out-of-scope guard, and backward-compat tests proving the unscoped path omits the new kwarg. Gap: no end-to-end DB test exercising update_conversation_scope's actual SQL/owner-scoping, though that's consistent with the repo's mock-boundary testing style. |
| plan_impl_fidelity | 5/10 | Not visible (source-plan.md not present) — expected for cell XO (plan=NONE). Only source-exploration.md exists; notably the implementation did address every risk the exploration enumerated, including the transcript-bypass edge case. |

---

PR #318, opened 2026-06-01
[bench-XK] review: address self-review findings

coleam00:
Fixes #279

**Source run-id:** `bd10cb3a-a2bb-4a97-abf2-0c6170a1e1e9`

### DIFF
--- a/app/backend/alembic/versions/0006_conversation_video_scope.py
+++ b/app/backend/alembic/versions/0006_conversation_video_scope.py
@@ -1,0 +1,34 @@
+"""Add video_ids scope to conversations.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-06-01
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        ALTER TABLE conversations
+        ADD COLUMN IF NOT EXISTS video_ids TEXT[]
+        """
+    )
+    op.execute(
+        "CREATE INDEX IF NOT EXISTS conversations_video_ids_idx ON conversations USING GIN (video_ids)"
+    )
+
+
+def downgrade() -> None:
+    op.execute("DROP INDEX IF EXISTS conversations_video_ids_idx")
+    op.execute("ALTER TABLE conversations DROP COLUMN IF EXISTS video_ids")
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -253,6 +253,7 @@ async def keyword_search(
     top_k: int,
     language: str = "english",
     allowed_source_types: list[str] | None = None,
+    allowed_video_ids: list[str] | None = None,
 ) -> list[dict]:
     """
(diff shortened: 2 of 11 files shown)


coleam00:
**Benchmark Evaluator** — score: **54/70**

A strong, complete implementation of conversation-level video scoping. The fix is applied at the correct layer — the retrieval SQL — and threaded consistently through the retriever, tool executors, and message route following the codebase's existing allowed_source_types/is_member pattern, so it is a genuine root-cause solution rather than a band-aid. It handles the trickiest edge case the exploration flagged (distinguishing NULL 'search everything' from empty 'search nothing') correctly on both create and query paths, ships a proper Alembic migration with a GIN index, and adds targeted tests plus honest updates to existing ones.

Weaknesses are modest. The frontend scope picker is a large inline-styled modal that, while consistent with the already inline-heavy Sidebar, drifts from the stated Tailwind convention. Test coverage stops short of the messages.py whitelist-building branch and the PATCH route's mixed update logic. The UI only lets users set scope at creation time even though the backend now supports editing it. Plan fidelity is unscored because this is the no-plan baseline cell. Overall a clean, well-scoped feature PR that solves the issue as written.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers the full feature end-to-end: pick a subset of videos at conversation creation, scope is persisted on the conversation row, every search/transcript tool is restricted to those ids, and an empty/NULL selection falls back to the whole library exactly as the issue asks. Scope 'sticks' because it is read from the DB on each message. Minor gap: scope is only settable at creation in the UI (no edit-after-create surface) though the backend PATCH supports it. |
| root_cause | 9/10 | Filtering is applied at the true source — the SQL retrieval queries — and threaded cleanly through retriever_hybrid → tools → messages route, mirroring the existing allowed_source_types/is_member pattern. This is a structurally correct fix, not a post-hoc filter on results. |
| scope_discipline | 8/10 | Every change is on-topic: migration, repository, retriever, tools, routes, frontend picker, and tests. Touches protected files (conversations.py, messages.py, repository.py) but only as strictly required and preserving owner scoping. The SELECT * → explicit-column change is justified to surface video_ids consistently. |
| subtle_correctness | 8/10 | Carefully distinguishes NULL/empty (search-all) from a populated scope via the (IS NULL OR array_length IS NULL OR = ANY) guard, and both create and messages paths coerce empty lists to None consistently. Deleted/re-ingested videos degrade to fewer results rather than erroring. Small inconsistency: create_conversation can persist an empty [] while the message path treats it as no-scope, harmless but slightly untidy. |
| code_quality | 7/10 | Backend is clean, idiomatic, and matches existing conventions; ConversationRename→ConversationUpdate is a sensible generalization. The frontend scope-picker modal is ~150 lines of inline styles rather than Tailwind — but that matches the surrounding Sidebar.tsx, which is already inline-style heavy, so it is consistent if not ideal. |
| test_discipline | 8/10 | Adds a focused test_video_scope.py covering repository SQL filtering, NULL handling, create_conversation, per-tool threading, execute_tool dispatch, and retrieve_hybrid threading; existing tests were updated to reflect new kwargs legitimately (not gamed). Gap: no test for the messages.py whitelist-building branch or the PATCH route's mixed title/video_ids logic. |
| plan_impl_fidelity | 5/10 | not visible (source-plan.md not present) — this is the XK no-plan baseline cell. The exploration artifact is present and the implementation tracks its recommended approach and edge cases very closely, but with no plan to compare against, scored neutral per instructions. |

---

PR #320, opened 2026-06-01
[bench-XK] impl: address issue

### LINKED ISSUE #295: Channel sync miscounts skipped videos as new and hides failed runs

coleam00:
Fixes #295

**Source run-id:** `54c8f57f-dfe1-4c86-9fbb-db082679f234`

### DIFF
--- a/app/backend/routes/channels.py
+++ b/app/backend/routes/channels.py
@@ -156,7 +156,6 @@ async def sync_channel(limit: int | None = None, force: bool = False) -> SyncRes
             existing = await repo.get_video_by_youtube_id(youtube_video_id)
             if existing is not None and not force:
                 logger.info("Video %s already ingested, skipping", youtube_video_id)
-                videos_new += 1
                 await repo.update_sync_video_status(
                     video_id=sync_video_record["id"],
                     status="ingested",
--- a/app/backend/tests/test_channel_sync.py
+++ b/app/backend/tests/test_channel_sync.py
@@ -145,7 +145,7 @@ def make_mock_transcript(text):
 async def test_sync_channel_idempotent_skips_existing_videos():
     """
     If a video is already in the DB (matched by youtube_video_id in URL),
-    it is skipped and counted as 'new' (already ingested = already counted).
+    it is skipped and must NOT be counted as 'new'.
     """
     # Pre-ingest a video so get_video_by_youtube_id returns it
     await repository.create_video(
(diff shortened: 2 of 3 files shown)


coleam00:

An excellent, tightly-scoped fix. The PR identifies that a single erroneous `videos_new += 1` in the already-ingested skip branch was the common root cause of both reported problems: it inflated the 'new videos' count, and because the run-status heuristic (`failed if videos_error > 0 and videos_new == 0`) keyed off that count, it also masked total-ingestion-failure runs as 'completed'. Removing the one line corrects both behaviors without touching the status logic, which was already correct — a genuine root-cause fix rather than a band-aid. Test discipline is strong: a new regression file covers both issue scenarios end-to-end with fully-mocked dependencies and asserts on the persisted aggregates, and the pre-existing idempotency test was honestly updated from asserting the buggy count to the correct one. The only weaknesses are minor: partial-failure runs (a mix of new successes and new failures) still report 'completed', which is left implicit, and plan-implementation fidelity cannot be scored because this is the no-plan baseline cell. Overall a high-quality, production-grade change.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly addresses both reported defects: skipped-existing videos no longer inflate the 'new' count, and because the run-status logic already keys off `videos_new == 0`, a run where all genuinely-new videos fail now correctly surfaces as 'failed'. |
| root_cause | 9/10 | True root cause. The single miscount (`videos_new += 1` in the skip branch) was the source of BOTH symptoms — it polluted the count AND defeated the existing `failed if videos_error > 0 and videos_new == 0` status check. Removing that one line fixes both without adding new status logic; the implementer correctly recognized the status check was already sound. |
| scope_discipline | 10/10 | Maximally surgical: a single production-line deletion in channels.py, plus a stale-assertion fix and a new regression test file. No unrelated churn. |
| subtle_correctness | 8/10 | Reasoning about the count/status interplay is correct and async handling is untouched. Partial-failure runs (some new succeed, some fail) still report 'completed', which matches the issue's intent but is an unstated edge; no foot-guns introduced. |
| code_quality | 9/10 | The production change is a clean deletion that leaves idiomatic code. Tests are readable, well-commented, use a helper, and assert on both the HTTP response and the captured `update_sync_run` args. |
| test_discipline | 9/10 | Added a dedicated regression file with one test per issue scenario (skip-not-counted, all-new-fail→failed). The modified existing test was asserting the buggy behavior; updating its assertion and docstring to the correct expectation is a legitimate fix, not gaming. |
| plan_impl_fidelity | 5/10 | not visible (source-plan.md not present) — this is the plan=NONE baseline cell, so there is no plan to compare against. |

---

PR #321, opened 2026-06-01
[bench-XO] fix(channels): count only newly-ingested videos, surface failed runs

### LINKED ISSUE #295: Channel sync miscounts skipped videos as new and hides failed runs

coleam00:
Fixes #295

**Source run-id:** `9aa62d5d-9860-4751-9b33-86ec3b92b8db`

### DIFF
--- a/app/backend/routes/channels.py
+++ b/app/backend/routes/channels.py
@@ -134,6 +134,7 @@ async def sync_channel(limit: int | None = None, force: bool = False) -> SyncRes
     videos_total = len(all_video_ids)
     videos_new = 0
     videos_error = 0
+    videos_skipped = 0
 
     logger.info(
         "Channel %s has %d videos (type=%s)",
@@ -156,7 +157,11 @@ async def sync_channel(limit: int | None = None, force: bool = False) -> SyncRes
             existing = await repo.get_video_by_youtube_id(youtube_video_id)
             if existing is not None and not force:
                 logger.info("Video %s already ingested, skipping", youtube_video_id)
-                videos_new += 1
+                # Already in the DB — not newly ingested this run, so it must not
+                # count toward videos_new. Inflating videos_new here both misreports
+                # the run as having ingested the whole channel and masks failed runs
+                # (the videos_new == 0 status check below would never fire). See #295.
+                videos_skipped += 1
                 await repo.update_sync_video_status(
                     video_id=sync_video_record["id"],
                     status="ingested",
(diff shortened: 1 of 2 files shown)


coleam00:

A tight, high-quality fix for a subtle two-part bug. The author correctly diagnosed that a single mislabeled counter (videos_new += 1 on the skip path) was the shared root cause of both the inflated 'new videos' count and the hidden failed runs, and corrected it by introducing videos_skipped without disturbing the already-sound status condition. Scope is exemplary — no needless migration (the existing constraint already permits 'ingested' for skipped rows) and no unrelated edits. The regression tests are the standout: they cover each bug independently and, importantly, the interaction case that masks outages, while respecting the repo's mocked-boundary testing rules and avoiding any gaming of the pre-existing skipped suite. Minor notes: videos_skipped is only logged, not persisted (no column, by design), and the partial-failure edge case remains 'completed' (intentionally out of scope). Plan fidelity is unscorable since this is the no-plan baseline cell, capping the total at 60/70 purely due to the absent planning artifact rather than any implementation weakness.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes both reported bugs: skipped (already-ingested) videos now increment videos_skipped instead of videos_new, and because videos_new is no longer inflated the existing 'failed if videos_error>0 and videos_new==0' check now fires correctly when all genuinely-new videos fail. Matches the issue's expected behavior exactly. |
| root_cause | 9/10 | Identifies the true single root cause — the erroneous `videos_new += 1` on the skip path — which was simultaneously inflating the count AND silently disabling the failed-run detection. One structurally-correct change fixes both symptoms; no band-aid. |
| scope_discipline | 10/10 | Surgical: only the skip-path counter and two log lines in channels.py, plus a new dedicated test file. No schema migration, no unrelated refactors, no churn in repository.py — correctly recognizing no new DB column was needed. |
| subtle_correctness | 9/10 | Status condition is left untouched and now works as intended; force=True and partial-failure paths are correctly unaffected. Skipped rows keep the valid 'ingested' status (no constraint violation). videos_skipped is logged but not persisted (no column) — acceptable and migration-avoiding. Async mocking is consistent. |
| code_quality | 9/10 | Clean, idiomatic edit with a clear explanatory comment tying the change to the masked-failure interaction and referencing #295. New counter is consistently threaded into the completion log line. Reads like the surrounding code. |
| test_discipline | 9/10 | Two meaningful regression tests, one per bug, including the critical interaction (a skip plus an all-new-fail run → 'failed'). Boundaries are mocked per CLAUDE.md guidance (no Postgres/Supadata/OpenRouter), placed in a separate module rather than gaming the skipped existing suite, and assert both the response and the persisted update_sync_run aggregate. |
| plan_impl_fidelity | 5/10 | not visible (source-plan.md not present) — this is the XO cell (plan=NONE), so no planning artifact exists on disk. The exploration artifact is present and the implementation aligns with it, but per instructions a missing plan scores 5. |

---

PR #322, opened 2026-06-01
[bench-XK] review: fix custom date parsing to use local time instead of UTC

coleam00:
Fixes #294

**Source run-id:** `9da4a653-11e3-443f-93de-b35f1cbc8c20`

### DIFF
--- a/app/backend/alembic/versions/0006_add_sources_gin_index.py
+++ b/app/backend/alembic/versions/0006_add_sources_gin_index.py
@@ -1,0 +1,31 @@
+"""Add GIN index on messages.sources for fast video filtering.
+
+Supports the conversation search by video feature (issue #294).
+A GIN index lets `@> jsonb_build_array(jsonb_build_object('video_id', ...))`
+run efficiently even with large message tables.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-06-01
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "CREATE INDEX IF NOT EXISTS messages_sources_gin_idx ON messages USING GIN (sources)"
+    )
+
+
+def downgrade() -> None:
+    op.execute("DROP INDEX IF EXISTS messages_sources_gin_idx")
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -526,6 +526,69 @@ async def search_conversations_by_title(user_id: str, query: str, limit: int = 2
     return [dict(r) for r in rows]
 
 
+async def search_conversations(
+    user_id: str,
+    *,
+    query: str | None = None,
+    date_from: str | None = None,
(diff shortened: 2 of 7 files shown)


coleam00:
**Benchmark Evaluator** — score: **48/70**

A genuinely solid feature PR for a no-plan baseline cell. It fully implements the requested capability — date-range and video filters that compose with text search and return newest-first results — and does so with the right architecture: server-side parameterized filtering replacing the old client-side substring filter, a GIN index for the JSONB video lookup, and reuse of the existing videos endpoint. The backend SQL is clean, injection-safe, and keeps ownership scoping intact even while editing a protected route file. The self-review pass added real value by fixing a local-vs-UTC date-parsing bug.

Weaknesses are concentrated in verification and timezone rigor. The added tests only confirm that query params reach the repository functions; they never exercise the actual filtering logic or the date math that carried the bug, so the meaningful behavior is untested. The timezone handling is also still half-finished: the client now computes local time correctly but transmits it without an offset to a TIMESTAMPTZ column, leaving the cross-zone ambiguity the exploration explicitly called out. Inline styling also drifts from the documented Tailwind convention. Net: correct and well-shaped feature work, undercut by shallow tests and an unresolved timezone edge case.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers exactly what the issue asked: server-side date-range filtering (presets + custom from/to), a video selector, all composable with the existing text search, ordered updated_at DESC (newest-first). The video filter correctly traverses the indirect link through messages.sources JSONB. |
| root_cause | 8/10 | Structurally sound: moves filtering from client-side JS to a parameterized server-side dynamic query, adds a GIN index to keep the JSONB video containment fast, and reuses the existing /api/videos endpoint for the dropdown. Not a band-aid — it addresses the scalability root the exploration flagged. |
| scope_discipline | 8/10 | Touches only relevant files (repository, route, hook, api wrapper, Sidebar, one migration, one test). Edits the protected conversations.py, but the change is strictly additive optional query params that preserve user_id scoping, and the route necessarily needs them. PR title oddly describes only the self-review fix rather than the feature, but no unrelated churn. |
| subtle_correctness | 6/10 | Mostly careful — startOfWeek/endOfDay roll-over math is correct via JS Date normalization, the EXISTS subquery stays user-scoped, ranges are inclusive, and the self-review caught the new Date('YYYY-MM-DD') UTC-shift bug. But toIsoLocal strips the timezone offset, so a local timestamp is sent to a TIMESTAMPTZ column and reinterpreted in the server's zone — the exact ambiguity the exploration warned about remains unresolved. |
| code_quality | 7/10 | Backend is clean and well-documented with sensible incremental param-index building; date helpers are small and well-named. Heavy use of static inline style={{...}} on the new selects/inputs violates the CLAUDE.md Tailwind-only rule, though it matches the surrounding Sidebar's existing inline-style idiom. |
| test_discipline | 5/10 | Adds hermetic tests, but they only mock the repository and verify the route forwards query params and branches between list/search. They cover the plumbing, not the substance — no test exercises the real SQL filtering/date casting/JSONB containment, nor the frontend computeDateRange/week-math where the UTC bug actually lived. |
| plan_impl_fidelity | 5/10 | Not visible (source-plan.md not present) — this is the no-plan baseline cell (XK, plan=NONE), so there is no plan to compare against; scored neutral per instructions. |

---

PR #323, opened 2026-06-01
[bench-XO] feat: filter conversation history by date range and video

coleam00:
Fixes #294

**Source run-id:** `f3afef28-f844-463f-a580-a8a19b47c1ec`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -35,6 +35,22 @@ def _now() -> datetime:
     return datetime.now(UTC)
 
 
+def _parse_dt(value: datetime | str) -> datetime:
+    """Coerce an ISO-8601 string (or passthrough datetime) to an aware datetime.
+
+    asyncpg binds TIMESTAMPTZ parameters as datetime objects — passing a raw
+    string raises DataError at bind-time (see _now). Frontend date-range
+    filters arrive as ISO-8601 strings (often with a trailing 'Z'), so
+    normalise them here before they hit a `WHERE updated_at >= $n` comparison.
+    """
+    if isinstance(value, datetime):
+        return value
+    text = value.strip()
+    if text.endswith("Z"):
+        text = text[:-1] + "+00:00"
+    return datetime.fromisoformat(text)
+
+
 def _acquire() -> asyncpg.pool.PoolAcquireContext:
     """Return the pool's acquire context manager.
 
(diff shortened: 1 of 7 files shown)


coleam00:
**Benchmark Evaluator** — score: **55/70**

A strong, focused implementation of issue #294. It fully solves the stated problem — date-range and video filters that combine with text search and preserve newest-first ordering — and does so at the correct architectural layer: parameterized, owner-scope-preserving SQL in repository.py, a clean api.ts contract, and graceful frontend wiring. Subtle correctness is a clear strength: ISO/timezone handling, NULL-safe JSONB containment, dynamic placeholder indexing, and a stale-response guard are all handled deliberately. Code is readable and well-commented, and stays within scope.

The main weakness is test depth: the tests verify that filters are parsed and forwarded but stop short of exercising the actual filtering logic (WHERE-clause assembly, _parse_dt, the @> video query), which is where the real risk lives — an integration test against the snapshot DB would have closed this gap. Inline styles technically contravene the Tailwind-only convention, though they match the existing Sidebar. Plan-fidelity is not scorable since this is the no-plan baseline cell. Total: 55/70.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers every part of the ask: custom date range, 'this week'/'last month' presets, a video dropdown, combination with the existing text search, and newest-first ordering (ORDER BY c.updated_at DESC preserved). Only minor gap is that the 'q' server param exists but text matching is still done client-side, which is acceptable and consistent. |
| root_cause | 9/10 | Filtering is pushed into the SQL layer in repository.py where it belongs, built as additive parameterized WHERE conditions on top of the mandatory owner scope. This is the structurally correct place, not a client-side band-aid that would break pagination/ordering. |
| scope_discipline | 9/10 | All 7 changed files are directly relevant (repository, route, api client, hook, Sidebar, plus two new test files). No drift into unrelated code, no opportunistic refactors, owner-scope invariant untouched. |
| subtle_correctness | 9/10 | Handles ISO 'Z' coercion to aware datetimes, UTC day-boundary conversion for inclusive ranges, NULL-safe JSONB @> containment, dynamic placeholder numbering, a fetch-ID guard against stale responses, and graceful video-dropdown degradation on getVideos() failure. Well-reasoned about edge cases. |
| code_quality | 8/10 | Clear naming, strong explanatory docstrings/comments, idiomatic asyncpg and React. Inline styles violate the Tailwind-only rule in CLAUDE.md, but Sidebar.tsx already uses inline styles throughout, so the change matches surrounding code rather than introducing a new pattern. |
| test_discipline | 6/10 | Adds genuine route-level (param forwarding) and frontend (query-string serialization) tests with no test-gaming. However it monkeypatches the repository, so the core logic — the dynamic WHERE construction, placeholder numbering, _parse_dt, and the JSONB video filter — is never exercised by a test. Meaningful but shallow on the riskiest code. |
| plan_impl_fidelity | 5/10 | not visible (source-plan.md not present) — this is the plan=NONE (XO) baseline cell, so there is no plan artifact to compare against. Only source-exploration.md exists. |

---

PR #324, opened 2026-06-10
[bench-FF] fix(integrations): Circle paid-membership check fails closed when active field absent

### LINKED ISSUE #243: fix(integrations): Circle paid-membership check fails open when active field is absent

coleam00:
Fixes #243

**Source run-id:** `6d5d09e1-d23b-49a2-9f3d-e8d06919bb7c`

Produced by the mixed-provider routing benchmark (Fable 5 workshop dress rehearsal).

### DIFF
--- a/app/backend/integrations/circle.py
+++ b/app/backend/integrations/circle.py
@@ -64,8 +64,10 @@ async def verify_paid_member(email: str) -> bool:
             member = _extract_member(r.json())
             if member is None:
                 return False
-            if not member.get("active", True):
-                # Inactive members lose access immediately.
+            if not member.get("active", False):
+                # Inactive members lose access immediately. A missing `active` field
+                # also denies: this is a paid-access gate and the contract is
+                # fail-closed — no explicit active signal means no access.
                 return False
             member_id = member.get("id")
             if not member_id:
--- a/app/backend/tests/test_circle.py
+++ b/app/backend/tests/test_circle.py
@@ -7,6 +7,7 @@ The function MUST never raise — it always returns a bool. We exercise:
 - non-200 from access_groups (fail-closed)
 - timeout (fail-closed)
 - inactive member
+- member with no `active` field (fail-closed)
 - member exists but NOT in paid access group
 - empty / malformed config
 - malformed response shape
@@ -107,6 +108,30 @@ async def test_inactive_member_returns_false():
     )
 
     assert await circle.verify_paid_member("inactive@example.com") is False
+
+
+@respx.mock
+async def test_member_missing_active_field_fails_closed():
+    """No `active` key on the member object must deny (fail-closed paid gate).
+
+    The access_groups endpoint is deliberately NOT mocked: a correct fix
+    short-circuits before step 2, so respx would raise on any call to it.
+    """
+    respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(
+        return_value=httpx.Response(200, json={"id": 999})
+    )
+
+    assert await circle.verify_paid_member("unknown-status@example.com") is False
+
+
+@respx.mock
+async def test_records_wrapped_member_missing_active_fails_closed():
+    """Same as above for the `records`-wrapped search shape, which often omits status."""
+    respx.get("https://app.circle.so/api/admin/v2/community_members/search").mock(
+        return_value=httpx.Response(200, json={"records": [{"id": 999}]})
+    )
+
+    assert await circle.verify_paid_member("unknown-status@example.com") is False
 
 
 @respx.mock


coleam00:

An exemplary, surgical fix. The PR correctly diagnoses the fail-open default (`member.get('active', True)`) as the root cause, flips it to `False` so that a member object lacking an explicit active signal is denied, and preserves every other behavior — explicitly-active members still proceed, inactive members still deny, and the access-group step, `_extract_member`, timeout, and request flow are all untouched per the out-of-scope constraints. The reasoning shown in the comment connects the module's documented fail-closed contract to the bug, exactly the connection the issue was testing for.

Test discipline is strong: two regression tests cover both the direct and `records`-wrapped search shapes the issue flagged as status-omitting, the docstring bullet list is updated, and one test deliberately leaves the access_groups endpoint unmocked so respx enforces that a correct fix short-circuits before step 2. The implementation is a faithful, complete execution of the plan with no drift and no scope creep. The only nit is that missing-key and explicit-`active: False` now collapse into one denial branch, which is intentional and correct for a fail-closed gate.

## Scores

|---|---|---|
| solves_issue | 10/10 | Changes the default from True to False so a member object lacking an `active` key now denies, while explicitly-active members still proceed to the access-group check. Exactly satisfies all three acceptance criteria. |
| root_cause | 10/10 | Targets the precise fail-open default (`member.get('active', True)`) the issue identified, restoring the documented fail-closed contract rather than papering over a symptom. No alternate-key guessing, which would be unsound. |
| scope_discipline | 10/10 | Two files: a one-line logic change plus comment in circle.py and targeted tests. `_extract_member`, timeout, and two-step flow all untouched as the issue demanded. |
| subtle_correctness | 9/10 | Correctly reasons that absence must deny; one test deliberately leaves access_groups unmocked so respx enforces the step-2 short-circuit. Preserves existing active/inactive behavior. Minor: `active: False` and missing-key now collapse to one branch, but that is intended. |
| code_quality | 9/10 | Clear comment explaining the fail-closed rationale, idiomatic to the module, consistent with existing respx test patterns. Production-grade and readable. |
| test_discipline | 10/10 | Adds two meaningful regression tests (direct and `records`-wrapped missing-`active` shapes), updates the test-module docstring, and asserts short-circuit by not mocking step 2. No existing tests weakened to pass. |
| plan_impl_fidelity | 10/10 | Implementation matches the plan verbatim — same default change, same comment text, both planned tests added including the records-wrapped variant and the deliberate unmocked-access_groups assertion. Zero drift. |

---

coleam00:

An exemplary small fix. The implementation correctly reasons that this is a fail-closed paid-access ACL, so absence of an explicit active signal must deny, and applies the minimal change (`get("active", False)`) that flips the missing-key case without regressing explicit active/inactive members. Scope is tightly disciplined to the two files the issue and plan call out, and the out-of-scope boundaries (_extract_member, timeout, two-step flow) are respected.

The tests are the standout: two regression cases covering both the direct and records-wrapped response shapes, with the access_groups endpoint deliberately left unmocked so respx fails if the gate doesn't short-circuit before step 2 — a genuine behavioral assertion, not a checkbox. Plan-to-implementation fidelity is essentially perfect, down to the comment wording. The only reason any dimension sits below 10 is that the change is trivial enough that there is little room to demonstrate depth beyond what was required; no real weaknesses are present.

## Scores

|---|---|---|
| solves_issue | 10/10 | Changes the fail-open default `member.get("active", True)` to `member.get("active", False)`, satisfying all three acceptance criteria exactly: missing active denies, explicit active proceeds to the unchanged group check, explicit inactive denies. |
| root_cause | 10/10 | Targets the precise default-value bug rather than papering over it; restores the documented fail-closed contract with the minimal correct change and leaves the rest of the two-step flow intact. |
| scope_discipline | 10/10 | Surgical: one logic line plus comment in circle.py and two regression tests in test_circle.py. No drift into _extract_member, timeout, request flow, or unrelated files. |
| subtle_correctness | 9/10 | Correctly handles both direct and records-wrapped shapes, and a null/falsy active also denies. The unmocked access_groups endpoint in the test cleverly proves the short-circuit before step 2. No foot-guns introduced. |
| code_quality | 9/10 | Idiomatic, readable, and the inline comment now states the fail-closed rationale clearly. Matches the existing respx/async test style. |
| test_discipline | 10/10 | Two meaningful, targeted regression tests for the exact bug, plus the docstring bullet update. Leaving access_groups unmocked turns the test into a real assertion that the gate short-circuits. No existing tests gamed. |
| plan_impl_fidelity | 10/10 | The diff matches the plan verbatim — same one-line default change, identical comment text, both planned tests (direct + records-wrapped) added, docstring updated, nothing extra. Faithful execution. |

---

PR #325, opened 2026-06-10
[bench-OO] fix(web): stop stream and skip onComplete on mid-stream error

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `830e0835-fb24-44d7-aa12-7c95642364e5`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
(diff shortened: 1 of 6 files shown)


coleam00:

An excellent, textbook fix. The implementation correctly diagnoses and repairs the nested-loop control-flow bug exactly as the issue and the plan describe: it labels the outer `while`, breaks it from the error branch so the reader stops promptly, and guards `onComplete` so a failed stream is never committed as success — all while preserving the finally-cleanup-then-throw ordering the React state resets depend on. The accompanying regression test is genuinely meaningful: it proves both that onComplete is skipped and that post-error tokens are not drained, which is the precise behavior that distinguishes the fix from the bug. Plan-to-impl fidelity is perfect.

The only real weakness is scope: the PR diff includes four unrelated benchmark workflow YAML files (FF/FK/FO/OF) that are benchmark-harness scaffolding rather than part of the issue fix, inflating an otherwise ~40-line change to 668 additions. This appears to be an artifact of the benchmarking process rather than the implementer overreaching into the application code, since the code change itself is impeccably surgical — but it still lands in the PR diff and pulls scope_discipline down. Otherwise this is production-grade work.

## Scores

|---|---|---|
| solves_issue | 10/10 | Hits all three acceptance criteria exactly: labeled `break streamLoop` stops the reader promptly on a `{"error"}` payload, `if (!streamError)` guards onComplete so partial text is never committed, and the post-finally `throw streamError` still surfaces the failure with finally running once. |
| root_cause | 10/10 | Addresses the actual nested-loop control-flow defect (inner `break` only exiting the `for`) rather than papering over it; labels the outer `while` and breaks it, which is the structurally correct fix the issue explicitly calls out. |
| scope_discipline | 7/10 | The code change is perfectly surgical — two edits to the hook plus one test. However the PR diff also carries four unrelated benchmark workflow YAML files (FF/FK/FO/OF), benchmark-harness scaffolding that has nothing to do with issue #244 and inflates the diff to 668 additions. |
| subtle_correctness | 10/10 | Preserves the finally-then-throw ordering the author relied on for React state resets, avoids the naive throw-inside-branch foot-gun, and won't double-run cleanup. No async/type issues introduced. |
| code_quality | 9/10 | Idiomatic, minimal, with a clear comment explaining the guard. The `streamLoop:` label is the standard JS way to break a nested loop. Reads cleanly and matches surrounding style. |
| test_discipline | 9/10 | Adds a focused regression test reusing existing helpers (makeSseStream/renderHook) that asserts onComplete is not called, the promise rejects with the error message, the post-error token is NOT accumulated (proving prompt loop exit), and isStreaming resets — directly distinguishing fixed vs. buggy behavior. |
| plan_impl_fidelity | 10/10 | The implementation executes source-plan.md exactly: label the outer loop, `break streamLoop` in the error branch, guard onComplete with `if (!streamError)`, leave finally and post-finally throw untouched, and add the prescribed regression test. Zero drift. |

---

PR #326, opened 2026-06-10
[bench-FO] fix(web): stop stream reader and skip onComplete on mid-stream error

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `1232f543-344c-4d6e-8f7e-b28b0c9d298b`

### DIFF
--- a/app/frontend/src/hooks/useStreamingResponse.test.ts
+++ b/app/frontend/src/hooks/useStreamingResponse.test.ts
@@ -614,3 +614,139 @@ describe('sources event — hook state via renderHook', () => {
     expect(onComplete).toHaveBeenCalledWith(expect.objectContaining({ sources: [] }));
   });
 });
+
+describe('mid-stream error payload — issue #244', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('rejects and never calls onComplete on a {"error"} payload', async () => {
+    const sseChunks = [
+      'data: "Partial answer"\n\n',
+      'data: {"error": "upstream exploded"}\n\n',
+      // These must never be consumed/committed once the error is seen
+      'data: " more text"\n\n',
+      'data: [DONE]\n\n',
+    ];
+
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        body: makeSseStream(sseChunks),
+      }),
+    );
+
+    const onComplete = vi.fn();
+    const { result } = renderHook(() => useStreamingResponse('conv-1'));
+
+    await act(async () => {
+      await expect(result.current.startStream('conv-1', 'hi', onComplete)).rejects.toThrow(
+        'upstream exploded',
+      );
+    });
+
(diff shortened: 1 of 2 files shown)


coleam00:

An exemplary minimal fix. The diff targets the true root cause — the inner `break` escaped only the `for` loop while the outer `while` kept draining to `done` and then fired `onComplete` with partial text — by labeling the outer reader loop and breaking it directly, then guarding `onComplete` behind `if (!streamError)`. It deliberately preserves the `try/finally` structure and the post-`finally` throw, sidestepping the exact trap the issue flagged (throwing inside the loop would disturb finally/cleanup ordering and the React 18 batching). Scope is tightly disciplined to the two files and leaves all explicitly out-of-scope branches untouched.

Test discipline is strong: beyond the obvious 'onComplete not called / promise rejects' case, the suite includes a hand-rolled reader that asserts `read()` is invoked exactly twice, which actually distinguishes a real loop-exit from a cosmetic onComplete guard that still drains the stream — the precise failure mode the issue called out. The malformed-error-JSON fallback and finally-cleanup state-reset tests round it out. Implementation fidelity to the planning artifact is near-perfect. The only nit is that no `reader.cancel()` is issued, but the plan correctly reasons this is beyond the minimal acceptance criteria and exiting the loop already halts consumption.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly satisfies all three acceptance criteria: the labeled `break readLoop` stops the reader loop promptly on a `{"error"}` payload, `onComplete` is guarded behind `if (!streamError)` so partial text is never committed, and `streamError` is still thrown after `finally`. |
| root_cause | 10/10 | Fixes the actual nested-loop control-flow defect — the inner `break` only escaped the `for` loop while the outer `while` kept draining. Labeling the outer loop addresses the structural cause rather than masking the symptom. |
| scope_discipline | 10/10 | Surgical: two source-line edits (label + labeled break) plus the onComplete guard, and one new test block. Explicitly leaves 401/429, SSE framing, and status-event handling untouched as the issue required. |
| subtle_correctness | 10/10 | Preserves the `try/finally` ordering the author relies on for React 18 batched state resets, avoids the throw-inside-loop trap the issue warned about, keeps the malformed-JSON default-message path, and the guard is correct belt-and-braces with the break. |
| code_quality | 9/10 | Idiomatic, minimal, and well-commented with an issue reference explaining the guard. Labeled break is the canonical readable solution here; nothing sloppy. |
| test_discipline | 10/10 | Four meaningful, targeted tests: rejection-without-onComplete, a hand-rolled reader asserting exactly two `read()` calls (proves prompt stop, not just guarded commit), finally-cleanup state resets, and malformed-error-JSON fallback. No existing tests gamed. |
| plan_impl_fidelity | 10/10 | The diff executes the plan essentially line-for-line: labeled `readLoop`, `break readLoop` in the error branch, the `if (!streamError)` guard with the same comment, and all four planned tests including the read-count assertion. Zero drift. |

---

PR #327, opened 2026-06-10

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `5c015562-0f91-4db4-966a-fbb150e4652c`

### DIFF
--- a/app/frontend/src/hooks/useStreamingResponse.test.ts
+++ b/app/frontend/src/hooks/useStreamingResponse.test.ts
@@ -537,6 +537,113 @@ describe('conversationId reset — state clears on navigation', () => {
   });
 });
 
+describe('mid-stream error event', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('rejects without calling onComplete when an error payload arrives after partial tokens', async () => {
+    const sseChunks = [
+      'data: "Partial "\n\n',
+      'data: {"error": "upstream model failure"}\n\n',
+      'data: "should never be appended"\n\n',
+      'data: [DONE]\n\n',
+    ];
+
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        body: makeSseStream(sseChunks),
+      }),
+    );
+
+    const onComplete = vi.fn();
+    const { result } = renderHook(() => useStreamingResponse('conv-1'));
+
+    await act(async () => {
+      await expect(result.current.startStream('conv-1', 'hi', onComplete)).rejects.toThrow(
+        /upstream model failure/,
+      );
+    });
+
+    expect(onComplete).not.toHaveBeenCalled();
+    // finally cleanup still ran exactly once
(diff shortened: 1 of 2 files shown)


coleam00:

An exemplary fix. The implementation correctly identifies the root cause (a bare `break` escaping only the inner for-loop while the outer `while` keeps draining the stream) and applies the minimal, structurally sound correction: label the reader loop, `break readLoop` from the error branch, and guard `onComplete` with `if (!streamError)`. This preserves the exact try/finally cleanup ordering and the post-finally throw the hook relies on, avoiding the naive throw-inside-loop trap the issue warned about. Scope is tightly disciplined, the code is readable with a clarifying comment, and the changes are confined to the two intended files.

Test discipline is the standout: beyond asserting onComplete isn't called and cleanup state resets once, the suite includes a pull-based ReadableStream that proves the reader stops promptly instead of draining to [DONE] — precisely the criterion a guard-only fix would fail — plus a malformed-payload fallback case. Plan-to-implementation fidelity is near-perfect, with the diff matching the detailed plan almost verbatim. The only quibble is the labeled-loop idiom being slightly uncommon JS, but it is the correct and well-documented tool here.

## Scores

|---|---|---|
| solves_issue | 10/10 | Hits all three acceptance criteria exactly: the labeled `break readLoop` stops the reader promptly, `onComplete` is guarded by `if (!streamError)` so it never fires on error, and the post-finally `throw streamError` still surfaces the failure with the `finally` running once. |
| root_cause | 10/10 | Fixes the actual control-flow defect — the bare `break` only escaping the inner for-loop — rather than papering over it. Labeling the outer loop is the structurally correct minimal fix and avoids the naive throw-inside-loop that would break cleanup ordering. |
| scope_discipline | 10/10 | Touches only the two intended files; the source change is a 3-line surgical edit with no incidental refactoring, and 401/429/SSE-framing/status handling are all left untouched per the out-of-scope list. |
| subtle_correctness | 10/10 | Correctly reasons about nested-loop control flow and the promise/cleanup lifecycle: labeled break preserves the try/finally ordering, the guard prevents committing partial text, and no double-cleanup or drained-stream race is introduced. |
| code_quality | 9/10 | Readable, idiomatic, well-commented explaining why the rest of the stream is discarded; the labeled-loop is the clean canonical pattern. Minor: a label is slightly unusual JS, but it's the right tool and clearly commented. |
| test_discipline | 10/10 | Three meaningful, targeted tests: onComplete-not-called + cleanup assertions, a clever pull-based ReadableStream proving the reader stops promptly rather than draining, and a malformed-JSON fallback case. No existing tests gamed. |
| plan_impl_fidelity | 10/10 | The diff executes the plan line-for-line — labeled `readLoop`, `break readLoop`, `if (!streamError)` guard, unchanged finally/throw — and all three planned tests appear essentially verbatim, including the pull-based prompt-stop test. |

---

PR #328, opened 2026-06-10
[bench-OF] review: address self-review findings

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `a3e081d6-3117-404b-bcdb-b0f8f81653e3`

### DIFF
--- a/app/frontend/src/hooks/useStreamingResponse.test.ts
+++ b/app/frontend/src/hooks/useStreamingResponse.test.ts
@@ -614,3 +614,124 @@ describe('sources event — hook state via renderHook', () => {
     expect(onComplete).toHaveBeenCalledWith(expect.objectContaining({ sources: [] }));
   });
 });
+
+describe('mid-stream {"error"} payload — issue #244', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('does not call onComplete and throws the server error', async () => {
+    const sseChunks = [
+      'data: "Partial answer"\n\n',
+      'data: {"error":"upstream failed"}\n\n',
+      // Content after the error must not resurrect the stream as a success
+      'data: " more text"\n\n',
+      'data: [DONE]\n\n',
+    ];
+
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        body: makeSseStream(sseChunks),
+      }),
+    );
+
+    const onComplete = vi.fn();
+    const { result } = renderHook(() => useStreamingResponse('conv-1'));
+
+    await act(async () => {
+      await expect(result.current.startStream('conv-1', 'hi', onComplete)).rejects.toThrow(
+        'upstream failed',
+      );
+    });
+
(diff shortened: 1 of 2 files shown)


coleam00:

An exemplary, near-perfect PR. It identifies and fixes the true root cause — a nested-loop control-flow bug where a bare `break` exited only the inner `for` loop, letting the outer `while` drain the rest of a failed stream and fire `onComplete` with partial text before the error was thrown. The fix is minimal and precisely the one the issue's 'why this tests reasoning' note prescribes: label the outer loop and `break streamLoop`, then guard `onComplete` with `if (!streamError)`, leaving the `finally` cleanup and trailing throw untouched so ordering and single-run cleanup are preserved. Scope is tightly disciplined, touching only the one production hook and its test file.

Test discipline is the standout: beyond asserting onComplete-not-called and the thrown message, the suite proves the loop stops *promptly* by gating a post-error chunk behind a never-resolving promise (the old bug would hang on reader.read()), verifies all four streaming-state resets run via finally, and covers the unparseable-payload default-message branch. Plan-to-implementation fidelity is complete. The only reason any dimension is below 10 is the inherent (and justified) reliance on a labeled loop; there are no correctness or quality faults to dock.

## Scores

|---|---|---|
| solves_issue | 10/10 | Implements exactly the acceptance criteria: labels the outer loop and uses `break streamLoop` so the reader stops promptly on a `{"error"}` payload, guards `onComplete` with `if (!streamError)`, and leaves the trailing `throw streamError` after `finally` intact so the failure still surfaces and cleanup runs once. |
| root_cause | 10/10 | Fixes the actual nested-loop control-flow defect — the bare `break` that only exited the inner `for` — rather than papering over the symptom. The labeled break terminates the outer `while`, preventing further `reader.read()` draining, which is the structurally correct fix. |
| scope_discipline | 10/10 | Surgical: 3-line production change in the single relevant hook plus a focused test block. No 401/429, SSE-parser, or status-event code touched, exactly as the issue's out-of-scope list demanded. |
| subtle_correctness | 10/10 | Correctly preserves the throw-after-finally ordering the author relies on for React state resets and avoids the naive in-branch throw that would double-run or skip cleanup. The 'no-drain' test even gates a post-error chunk behind a never-resolving promise to prove the loop stops rather than blocking forever. |
| code_quality | 9/10 | Clean, idiomatic, well-commented (comments explain why the break exits the outer loop and why onComplete is guarded). Matches surrounding style. Minor: nothing to fault beyond the inherent reliance on a labeled loop, which is justified here. |
| test_discipline | 10/10 | Four meaningful, targeted tests: onComplete-not-called + error thrown, prompt-stop via an unreleased gate, finally state-reset verification, and the unparseable-payload default-message path. No existing tests gamed or weakened. |
| plan_impl_fidelity | 10/10 | The diff executes the plan verbatim — labeled `streamLoop`, `break streamLoop`, `if (!streamError)` guard, untouched finally/throw — and adds the exact four test cases the plan enumerated, including the optional malformed-payload test. |

---

PR #329, opened 2026-06-10

### LINKED ISSUE #244: fix(web): streaming onComplete fires with partial text after a mid-stream server error

coleam00:
Fixes #244

**Source run-id:** `3b31bfc9-0104-40a5-9d68-9efd16e17df6`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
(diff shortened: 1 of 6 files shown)


coleam00:

A clean, correct, high-fidelity fix. The implementer adopted precisely the issue-endorsed approach — label the outer read loop, break out of it on a mid-stream error payload, and guard onComplete with if(!streamError) — while leaving the finally cleanup and the trailing throw untouched, which preserves the once-only cleanup ordering and React 18 batching the author relies on. This is the genuine root-cause fix, not the tempting guard-only band-aid that would keep draining the stream. The regression suite is strong: it includes a 'stops promptly' test asserting the reader doesn't consume post-error chunks, which is exactly the behavior that separates a real fix from a superficial one, plus default-error-message and finally-once coverage. Plan-to-implementation fidelity is essentially perfect.

The only blemish is scope: the PR diff includes four unrelated benchmark-cell YAML files (FF, FK, FO, OF, ~620 lines) that are benchmark harness scaffolding rather than part of the streaming fix. They don't touch app code and are an artifact of how the benchmark branch is assembled, so they don't undermine correctness, but they do bloat the diff and cost a couple of scope-discipline points. Net: an excellent, surgical fix to the actual code with thoughtful tests.

## Scores

|---|---|---|
| solves_issue | 10/10 | Exactly addresses the stated problem: a mid-stream {"error"} payload now stops the read loop promptly (labeled break), onComplete is guarded behind !streamError so partial text is never committed, and streamError is still thrown so the caller surfaces the failure. All three acceptance criteria met. |
| root_cause | 10/10 | Fixes the true root cause — the inner break only exited the for-loop while the while(true) kept draining. The labeled break exits the outer loop too, so the stream stops instead of being papered over by only guarding onComplete (which would still drain). |
| scope_discipline | 7/10 | The app-code change (useStreamingResponse.ts + its test) is perfectly surgical, 6 lines + targeted tests. Dock for the diff carrying 4 unrelated benchmark-cell YAML files (~620 lines); these are benchmark harness plumbing swept onto the branch, not the issue fix, but they inflate the PR. |
| subtle_correctness | 9/10 | Correctly preserves the finally-runs-once semantics and cleanup ordering the issue warned about (no throw from inside try, so finally fires once, error thrown after cleanup). Handles the unparseable-error-payload default-message path. No foot-guns introduced. |
| code_quality | 9/10 | Idiomatic and minimal — labeled-loop break reads cleanly, matches existing code style, retains the explanatory comment. Production-grade. |
| test_discipline | 9/10 | Four meaningful regression tests including the critical 'stops promptly' test (pullCount < chunks.length) that distinguishes the real fix from the guard-only non-fix, plus finally-cleanup-once and default-message coverage. Tests target real behavior, not gamed. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan line-for-line: labeled while loop, break readLoop, guarded onComplete, finally/throw left untouched, and the exact four-test suite the plan specified. Zero drift. |

---

PR #330, opened 2026-06-10
[bench-OO] fix(rag): skip even-distribution for zero-duration segments

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `b0f51a39-b567-4919-a14f-590114ffd4eb`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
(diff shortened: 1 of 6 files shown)


coleam00:

An excellent, textbook fix. The implementer correctly identified that the even-distribution math is undefined when a segment has non-positive duration and guarded the branch with `duration > 0`, leaving each sub-chunk on its original [start_s, end_s] bounds. This is the precise root-cause reasoning the issue was testing for, avoiding both tempting band-aids (clamping step, faking a duration). The change is minimal and idiomatic, the comment explains the why, and the added regression test meaningfully exercises the zero-duration multi-sub-chunk path while the existing positive-duration test guards normal behavior. Plan-to-implementation fidelity is essentially perfect. The only blemish is scope: the PR carries four unrelated benchmark-harness YAML files alongside the two-file fix, which inflates the diff and slightly weakens scope discipline — though these are clearly an artifact of the benchmark experiment rather than the implementer adding unrelated logic to the codebase under test.

## Scores

|---|---|---|
| solves_issue | 10/10 | Adds the `duration > 0` guard so zero/negative-duration segments skip even-distribution and keep their original [start_s, end_s] bounds. Directly satisfies all three acceptance criteria: no all-equal collapse, positive-duration distribution preserved, fallback untouched. |
| root_cause | 10/10 | Targets the true root cause — the distribution math is meaningless when duration <= 0 — rather than the called-out wrong fixes (clamping step or inventing a synthetic duration). Correctly connects the chunker branch to the upstream zero-width final segment. |
| scope_discipline | 7/10 | The actual code fix is perfectly surgical: one guard condition in chunker.py plus one test. However the PR bundles four unrelated benchmark workflow YAMLs (FF/FK/FO/OF, ~620 lines), which are experiment scaffolding rather than part of issue #245 — diluting an otherwise clean diff. |
| subtle_correctness | 9/10 | Handles the exact edge case and also covers negative durations via `> 0`. The `end_seconds >= start_seconds` invariant is preserved, and sub-chunks retain their construction-time bounds. No async/type foot-guns introduced. |
| code_quality | 9/10 | Clear, well-scoped comment explaining why distribution is skipped; hoisting `duration` out of the branch is idiomatic and reads cleanly. Matches surrounding style. |
| test_discipline | 9/10 | Adds a focused regression test asserting every chunk keeps start==end==start_s for a zero-duration multi-sub-chunk segment, plus the end>=start invariant. Meaningful, not gamed. Minor dependence on HybridChunker actually splitting the text, but the assertions hold regardless. |
| plan_impl_fidelity | 10/10 | The implementation matches the source plan almost verbatim — same guard, same comment intent, same test name and assertions, same out-of-scope discipline. No drift between plan and diff. |

---

PR #331, opened 2026-06-10
[bench-FO] fix(rag): keep boundary timestamps for zero-duration segments

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `d850aeca-209a-4a78-a4a4-c1b096fe0793`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
+      Use Read, Grep, and Glob to explore the repository. Look for:
+      - Files and modules involved in the issue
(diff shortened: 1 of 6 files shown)


coleam00:

An exemplary fix. The Fable plan correctly diagnosed the two-file correctness property — that _parse_segments/Supadata can emit a zero-width final segment and that even-distribution is meaningless without real duration — and the Opus implementation executed that plan almost character-for-character. The code change is a 4-line indentation-and-guard wrap with an accurate comment, addressing the true root cause while sidestepping both tempting wrong fixes the issue calls out. Test discipline is excellent: three focused cases covering zero, negative, and positive duration, each first asserting a real multi-sub-chunk split occurred so the regression branch is genuinely exercised, with a shared fixture and pytest.approx for the contiguity checks.

The only blemish is that the PR diff carries four unrelated benchmark-*.yaml harness files alongside the fix, an artifact of the experimental branch rather than the implementer's intent — the substantive code change touches only chunker.py and its test file and is genuinely surgical. Plan/impl fidelity is near-perfect, making this a strong showcase for the Fable-plan/Opus-impl cell.

## Scores

|---|---|---|
| solves_issue | 10/10 | Guards the even-distribution loop behind `duration > 0`, so zero-width (end==start) segments that split into multiple sub-chunks keep their boundary timestamps instead of collapsing to a single instant. Satisfies all three acceptance criteria: no all-equal distributed timestamps, normal positive-duration distribution preserved, fallback path untouched. |
| root_cause | 10/10 | Hits the exact reasoning the issue rewards: recognizes even-distribution is meaningless when there is no real duration, and relies on the sub-chunks already being initialized to the source [start_s, end_s] boundary. Avoids both named wrong fixes (no invented minimum step, no pinning every chunk to start_s). |
| scope_discipline | 8/10 | The actual code change is perfectly surgical — only chunker.py (a 4-line guard) and the dedicated test file. The PR diff also bundles four new benchmark-*.yaml harness files (FF/FK/FO/OF), which are benchmark-infrastructure scaffolding unrelated to issue #245; these are an artifact of the experiment branch rather than implementer scope creep, but they do inflate the diff. |
| subtle_correctness | 10/10 | Correctly handles both zero AND negative duration in one guard, leaning on the pre-existing boundary initialization at append time so no else-branch is needed. No type, async, or off-by-one foot-guns; the explanatory comment captures the intent precisely. |
| code_quality | 10/10 | Minimal, idiomatic indentation-only wrap of the existing loop with a clear, accurate comment explaining the zero-width case. Reads like the surrounding code and matches the repo's commenting density. |
| test_discipline | 10/10 | Three targeted tests (zero-duration, negative-duration, positive-duration) sharing a single module-level _LONG_TEXT fixture sized to force a real multi-sub-chunk split. Each asserts len(result) > 1 first to prove the distribution branch is actually exercised, and the positive case uses pytest.approx for contiguity and monotonicity — no gaming. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan essentially verbatim: identical guard, the exact comment text, the same two files, and all three test names with the same `range(400)` fixture and assertion strategy described in source-plan.md. No drift in either direction. |

---

PR #332, opened 2026-06-10

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `0d7b1849-7791-46eb-9929-2176674436d0`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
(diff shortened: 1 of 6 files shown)


coleam00:

An excellent, textbook fix for the FF (plan=Fable, impl=Fable) cell. It identifies the true root cause — that even-distribution is meaningless for a zero/negative-duration segment — and resolves it by simply skipping redistribution so the sub-chunks retain the boundary they were assigned at append time, avoiding every trap the issue flagged (no step clamping, no invented durations, no special-case-that-still-pins-to-start_s). It even hardens against the negative-duration variant via `> 0` rather than `!= 0`. Tests are well-targeted and guard against vacuous passes, and the implementation tracks the plan almost line-for-line.

The only blemish is diff size: the PR bundles four new benchmark-cell YAML files (FF/FK/FO/OF) totaling ~620 lines, which dwarfs the 9-line code+comment fix. These are experiment-harness artifacts rather than genuine scope creep in the code change, so the deduction is small. Overall a high-quality, production-ready PR.

## Scores

|---|---|---|
| solves_issue | 10/10 | Guards the redistribution branch with `duration > 0`, so a zero-duration splitting segment keeps each sub-chunk on the original [start_s, end_s] boundary instead of collapsing to start_s. Exactly the acceptance criteria, and positive-duration distribution is preserved. |
| root_cause | 10/10 | True root cause: it recognizes the even-distribution is meaningless when duration<=0 and relies on the already-correct append-time boundary, rather than clamping step or inventing a fake duration (the tempting wrong fixes the issue called out). |
| scope_discipline | 8/10 | The production change is a 5-line surgical edit to one function plus targeted tests; _parse_segments and fallback untouched. The 4 new benchmark-*.yaml files inflate the diff to 669 additions, but those are benchmark-harness scaffolding emitted by the experiment, not real scope creep in the fix. |
| subtle_correctness | 10/10 | Uses `duration > 0` rather than `!= 0`, so malformed end<start segments also skip redistribution instead of producing reversed spans — a subtle case explicitly verified by a negative-duration test. |
| code_quality | 9/10 | Clean, idiomatic edit with an accurate explanatory comment tying the guard to _parse_segments' end==start fallback. Nothing to fault. |
| test_discipline | 10/10 | Three meaningful targeted tests: zero-duration, negative-duration, and a positive-duration regression lock. Each asserts len(result) > 1 first so a silently-merged fixture fails loudly rather than passing vacuously. |
| plan_impl_fidelity | 10/10 | The diff is a faithful, near-verbatim execution of source-plan.md — same production guard, same three test names and fixtures (range(1500)), same out-of-scope boundaries. No drift. |

---

PR #333, opened 2026-06-10

### LINKED ISSUE #245: fix(rag): chunk_video_timestamped collapses timestamps for zero-duration segments

coleam00:
Fixes #245

**Source run-id:** `590c031b-9a29-4913-9276-5eadec11e3c2`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
+      Use Read, Grep, and Glob to explore the repository. Look for:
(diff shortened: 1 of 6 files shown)


coleam00:

An excellent, high-fidelity fix. The implementer (Fable) precisely executed an Opus plan that already nailed the root cause: redistribute timestamps only when the segment has a strictly positive duration, otherwise keep the original [start_s, end_s] bounds each sub-chunk received at append time. This avoids both tempting wrong fixes (clamping step to a minimum, or special-casing duration==0 while still pinning every chunk to start_s). The two-file correctness property linking the chunker's distribution branch to _parse_segments' zero-width final segment is handled correctly, and the change respects every out-of-scope constraint. Tests are targeted and meaningful, covering zero, negative, and positive duration with real multi-chunk splits.

The only blemish is PR-level scope: the diff bundles four large benchmark cell YAML files (FF/FK/FO/OF) that have nothing to do with issue #245. These are benchmark-orchestration scaffolding pulled in from the worktree rather than the implementer inventing unrelated changes, and the actual fix remains a clean ~5-line edit, but they inflate the PR from a surgical change to 666 additions across 6 files. Setting that harness noise aside, the code, reasoning, and tests are production-grade.

## Scores

|---|---|---|
| solves_issue | 10/10 | Guards the redistribution branch on `duration > 0`, so a segment with `end_s <= start_s` that splits into multiple sub-chunks keeps its original [start, end] bounds instead of collapsing to a single instant. Meets every acceptance criterion: degenerate case fixed, positive-duration even-distribution preserved, fallback path untouched. |
| root_cause | 10/10 | Targets the actual reasoning trap the issue describes — when duration<=0 the proportional split is meaningless, so it skips distribution and leaves each sub-chunk on the boundary it already got at append time. No fake-duration clamp, no special-case that still pins everything to start_s. |
| scope_discipline | 7/10 | The code fix is genuinely surgical (one guard condition + a comment in chunker.py, plus a test file). However the PR diff also carries four ~155-line benchmark YAML files (FF/FK/FO/OF) unrelated to issue #245; these are benchmark-harness scaffolding swept in from the worktree rather than implementer sprawl, but they still dilute the PR. |
| subtle_correctness | 10/10 | Correctly handles both zero AND negative (float-rounding) duration via the single `duration > 0` test, preserves monotonicity for the normal path, and introduces no foot-guns. Computing `duration` once before the guard reads cleanly. |
| code_quality | 9/10 | Readable, idiomatic to the module, with an accurate comment explaining why zero/negative-duration segments keep original bounds. Minimal and well-placed; nothing to fault beyond it being a trivially small change. |
| test_discipline | 9/10 | Three meaningful, targeted regression tests: zero-duration collapse, negative-duration, and positive-duration-still-distributes (locking in the unchanged path). Text is padded to force a real multi-chunk split. Robust to single-chunk outcomes; no existing tests gamed. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan essentially verbatim — same guard rewrite, same comment update, and the planned zero-duration + positive-duration tests, plus the negative-duration case the plan also flagged. No drift. |

---

PR #335, opened 2026-06-10
[bench-OO] fix(messages): finalize sources identically on interrupted streams

coleam00:
Fixes #277

**Source run-id:** `6e3391b8-954c-4394-b484-37171d66b173`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
+      Use Read, Grep, and Glob to explore the repository. Look for:
+      - Files and modules involved in the issue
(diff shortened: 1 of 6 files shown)


coleam00:

An excellent, textbook fix. It correctly diagnoses that the dedup/collapse/cap/refusal-suppress pipeline was trapped inside the [DONE] branch, so an interrupted or errored stream persisted a never-finalized list that disagreed with the live view. The remedy — a single pure `_finalize_source_citations` helper invoked by both the live SSE path and the finally-block persist path, always derived from the same raw `tool_chunks_acc` — is the structurally correct root-cause fix and guarantees parity by construction. Code quality and docstrings are production-grade and idiomatic to this codebase, and the regression tests target the exact failure mode (no [DONE]) plus refusal and dedup/collapse edge cases. Implementation fidelity to the plan is near-perfect.

The only real knock is scope: the PR diff bundles four unrelated benchmark workflow YAML files (~620 lines) alongside the genuine fix, inflating the changeset well beyond issue #277. That's an artifact of the benchmark harness rather than a logic-scope violation in the application code itself, and the actual code edit remains tightly surgical. Minor nit: the helper mutates the input chunks' is_cited flag, which is idempotent here but slightly impure.

## Scores

|---|---|---|
| solves_issue | 10/10 | Directly fixes the reported divergence: the persist path now runs the raw accumulator through the same finalizer as the live `event: sources`, so an interrupted/errored stream (no [DONE]) saves deduped, video-collapsed, refusal-suppressed sources identical to what the user saw. |
| root_cause | 10/10 | Identifies the true root cause — finalization logic lived only inside the [DONE] branch while the finally-block persisted an un-finalized list — and fixes it structurally by extracting one pure helper shared by both call sites, deriving from the same `tool_chunks_acc`. Not a band-aid. |
| scope_discipline | 7/10 | The substantive code change is surgical: one new helper, two call sites rewired, the redundant `source_citations` accumulator deleted, one new test file. However the PR diff also bundles four unrelated benchmark workflow YAMLs (FF/FK/FO/OF, ~620 lines) that have nothing to do with issue #277. |
| subtle_correctness | 9/10 | Thoughtful about failure modes: handles no-[DONE] interruption, refusal->None, empty/no-chunk_id->None, and preserves the exact original ordering (dedup -> is_cited -> collapse -> cap -> refusal-suppress). Double invocation on the clean path is idempotent. Minor: the helper mutates input chunk dicts' is_cited, harmless here but a mild foot-gun. |
| code_quality | 9/10 | The pure helper is well-named, has an excellent docstring tracing issue #277 and the contract, and matches the codebase's dense explanatory-comment style. Reuses existing primitives (_collapse_by_video, extract_cited_chunk_ids, _is_refusal) cleanly. |
| test_discipline | 9/10 | Strong targeted regression coverage: integration tests drive the ASGI app through an interrupted stream that errors without [DONE] and assert persisted sources are deduped/collapsed/refusal-suppressed and equal to the shared finalizer, plus focused unit tests of the helper. A client-disconnect/CancelledError variant would round it out, but the error-payload path exercises the core bug. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — same helper signature and ordered behavior, both call sites rewired exactly as specified, the `source_citations` accumulator removed, and all four prescribed test cases (deduped/collapsed, matches-helper, interrupted-refusal->None, plus unit tests) present with the planned names. |

---

PR #336, opened 2026-06-10

coleam00:
Fixes #277

**Source run-id:** `e774a0f9-3eeb-48c4-b00b-68cb382fd25b`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -171,6 +171,7 @@ async def create_message(
     async def event_generator() -> AsyncGenerator[str, None]:
         full_response: list[str] = []
         final_text_buf: list[str] = []
+        persisted_sources: list[dict] | None = None
         # Two-tier citations (issue #176): strip `[c:<id>]` markers from the
         # stream; use them at [DONE] to flag is_cited on retrieved chunks.
         marker_stripper = CitationMarkerStripper()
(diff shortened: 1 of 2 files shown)


coleam00:

A strong, focused fix. The PR correctly diagnoses that the live `event: sources` emission and the persisted `sources` column were two independent computations that could disagree on interrupted/errored streams, and unifies them by capturing the exact emitted payload immediately after the live yield and persisting only that. The capture-after-yield placement is a subtle but correct exploitation of async-generator resume semantics: the line runs only once the consumer has accepted the sources chunk, so an aborted send leaves persistence at None — matching what the client actually saw. Scope is tight (backend-only, two files), the dead refusal re-derivation is cleanly removed, and the comments explain the non-obvious ordering.

The test suite is the highlight: four manual-generator tests cover clean completion, interruption at the sources yield, interruption after the event was delivered, and a mid-stream upstream error, with the interrupt-before-sources case serving as a real regression guard that the previous code would fail. Minor nitpicks only: the capture is a reference rather than a copy (safe here since source_citations isn't mutated afterward), and "client received the event" is really "send() succeeded," which is the best achievable signal at the protocol boundary. Plan-to-implementation fidelity is near perfect.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly addresses the reported divergence: it makes the persisted sources equal to what was emitted live, and persists None when the live sources event never reached the client (interrupt/error mid-stream). That is exactly the expected behavior the issue asked for. |
| root_cause | 9/10 | True root-cause fix: it removes the second, independent re-derivation in the finally block and ties persistence to the live emission via a single captured value, so the two computations can no longer drift. Not a band-aid. |
| scope_discipline | 10/10 | Surgical — backend-only, two files (one logic file plus its co-located test), no unrelated edits. Correctly leaves frontend, refusal patterns, and stream_chat untouched per the plan's out-of-scope section. |
| subtle_correctness | 9/10 | The capture-after-yield is a correct, deliberate use of async generator suspend/resume semantics: the line runs only after the consumer accepts the yielded chunk, so a failed send / aclose leaves it None. Captures a reference rather than a copy, but source_citations is not mutated afterward, so it's safe. |
| code_quality | 9/10 | Readable and idiomatic to the file; the explanatory comments clearly justify the non-obvious ordering, and the now-dead refusal_check_text re-derivation is cleanly removed. Production-grade. |
| test_discipline | 9/10 | Four targeted regression tests driving event_generator manually (clean completion, interrupt-at-sources-yield, sources-then-interrupt, mid-stream error). The interrupt-before-sources test is a genuine guard that fails against the old re-derivation logic. Tests assert real persisted-vs-emitted equality, not gamed. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan essentially verbatim — same persisted_sources variable, same after-yield placement, same finally-block replacement, and the four named tests all present with the described semantics. |

---

PR #337, opened 2026-06-10

coleam00:
Fixes #277

**Source run-id:** `f4aed9de-2266-4f89-b9c4-85ba159d941b`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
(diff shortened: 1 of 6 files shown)


coleam00:

A strong, root-cause fix for issue #277. The implementer correctly identified that the bug was persistence reading a computed closure variable (`source_citations`) rather than what was actually delivered to the client, and solved it by capturing a snapshot immediately after the sources SSE frame's yield resumes — code that only runs when the consumer successfully received the frame. This makes 'persisted == shown' hold by construction across every completion mode (clean, interrupted, errored, refusal, no-chunks), and removes the second, divergent refusal check that was the latent inconsistency. The async reasoning is careful and correct, the change is confined to the exact lines it needed to touch (respecting the protected auth sections of this file), and the three regression tests exercise precisely the failure modes the issue described.

The only blemish is that the PR diff bundles four unrelated benchmark workflow YAML files (FF/FK/FO/OF) alongside the two-file code change — these are benchmark-harness artifacts rather than fix scope, inflating the changed-file count, though the meaningful fix itself is exemplary and surgical. Plan-to-implementation fidelity is essentially perfect.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes the reported divergence: persisted sources now track exactly what was delivered live, so an interrupted/errored stream persists None instead of a longer, un-collapsed chip set. Covers the duplicate-chip and refusal-with-chips symptoms the issue describes. |
| root_cause | 10/10 | True structural fix — captures `emitted_sources` only after the sources `yield` resumes, making persistence track delivery by construction, and deletes the divergent second `_is_refusal` check in `finally` that was the actual source of inconsistency. Not a band-aid. |
| scope_discipline | 8/10 | The code change is surgical: only `messages.py`'s generator persistence logic and the test file are touched, respecting the protected auth/ownership sections. The 4 new benchmark-*.yaml files are benchmark-harness scaffolding unrelated to the fix, which adds noise to the diff but isn't fix-scope creep. |
| subtle_correctness | 9/10 | Correctly reasons about async generator semantics — post-yield code runs only if the consumer resumed after receiving the frame; `aclose()` at the yield leaves it None. Shallow copy is sound (dicts not mutated after), and the `asyncio.shield` persistence pattern is preserved untouched. |
| code_quality | 9/10 | Clear, well-placed comments explaining the invariant, idiomatic to the file's existing heavy-comment style, good naming (`emitted_sources`). Removes now-dead `refusal_check_text` derivation cleanly. |
| test_discipline | 9/10 | Three targeted regression tests: generator closed at the sources yield persists None (the core bug), upstream error without [DONE] persists None while keeping partial text, and happy-path persisted-equals-emitted with dedup/collapse assertions. They pin the exact contract and would fail before the fix. |
| plan_impl_fidelity | 10/10 | Implementation executes the plan step-for-step: the `emitted_sources` local, the post-yield capture, the `finally` simplification, the rewritten comment block, and all three planned tests appear exactly as specified, with out-of-scope items left untouched. |

---

PR #338, opened 2026-06-10
[bench-FO] impl: persist sources matching the live stream on interrupted answers

coleam00:
Fixes #277

**Source run-id:** `03b3d028-b713-439c-9f43-da2efb2e0017`

### DIFF
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -118,11 +118,12 @@ async def create_message(
 
     # 5. Set up tool plumbing. All retrieval happens inside the LLM loop via
     # tool calls — no pre-retrieval runs here. The executor closure collects
-    # every chunk returned by any tool call so the final SSE `sources` event
-    # lists exactly what the model actually read. The video_id whitelist is
-    # only consulted by the transcript tool (it guards against hallucinated
-    # ids); the search tools ignore it.
-    source_citations: list[dict] = []
+    # every chunk returned by any tool call into ``tool_chunks_acc`` (the raw
+    # accumulator) so the final SSE `sources` event lists exactly what the
+    # model actually read. The dedup/collapse/cap/refusal pipeline runs once,
+    # in ``_finalize_sources``, at [DONE]. The video_id whitelist is only
+    # consulted by the transcript tool (it guards against hallucinated ids);
+    # the search tools ignore it.
     tool_chunks_acc: list[dict] = []
     embedding_cache: dict[str, list[float]] = {}
     tools_param: list[dict] | None = None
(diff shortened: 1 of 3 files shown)


coleam00:

A high-quality, well-targeted fix. It correctly diagnoses the root cause as two divergent source-finalization paths over shared mutable state and eliminates the divergence by extracting one pure `_finalize_sources()` helper and persisting exactly the list emitted to the client (or None). The load-bearing detail — assigning `emitted_sources` only after the `event: sources` yield so a disconnect leaves it None — is handled correctly and clearly documented, and purity via `dict(tc)` avoids mutating the accumulator. Test coverage is excellent: unit tests pin the helper's pipeline semantics and integration tests reproduce the issue's interrupted/errored/refused scenarios plus an emitted-equals-persisted equality assertion.

Weaknesses are minor. The change lands in `routes/messages.py`, which CLAUDE.md/FACTORY_RULES marks as a protected path; the plan explicitly anticipated flagging this for human review in the PR body, but the PR body omits that callout (the diff itself stays cleanly within the sources pipeline, so the risk is low). `emitted_sources or None` is slightly redundant since the variable is only ever None or a non-empty list. The genuinely unsolvable TCP-buffering window (sources bytes flushed to the ASGI layer but lost on a dying socket) is correctly scoped out and acknowledged rather than papered over. Overall a clean, faithful, root-cause fix.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly addresses the reported bug: live and reloaded sources now agree by construction because the persisted row is exactly the list handed to the client (or None). Interrupted/errored/refused streams persist None, killing the duplicate/longer/declined-answer chips on reload. Regression tests cover the exact scenario in the issue. |
| root_cause | 10/10 | Fixes the true root cause — two divergent finalization paths over shared mutable `source_citations`. It unifies them into one pure `_finalize_sources()` and persists strictly what was emitted, making divergence structurally impossible rather than patching one symptom. |
| scope_discipline | 9/10 | Surgical: one route file plus two test files, strictly within the sources pipeline; auth/ownership/rate-limit/shield semantics untouched. Minor ding — messages.py is on CLAUDE.md's protected-paths list and the plan called for the PR body to flag it for human review, which the PR body omits. |
| subtle_correctness | 9/10 | Thoughtful about the hard parts: the load-bearing post-yield `emitted_sources = finalized` assignment that stays None on GeneratorExit/CancelledError, purity via `dict(tc)` so the accumulator is never mutated, falsy chunk_id skipping, and a unified final_text for both is_cited and the refusal gate. No async/await or type foot-guns introduced. |
| code_quality | 9/10 | Production-grade: clear docstring enumerating the pipeline, accurate inline comments explaining the disconnect window, idiomatic helper placement near `_collapse_by_video`, correct type hints. `emitted_sources or None` is mildly redundant (it's never the empty list) but defensive and harmless. |
| test_discipline | 9/10 | Strong, meaningful tests: 8 unit tests pinning `_finalize_sources` (dedup, falsy ids, collapse+segment_count, is_cited, cap, refusal, purity) plus 4 integration tests via the real route harness covering interrupted error payload, mid-stream exception, refusal-cut-off, and the emitted==persisted equality invariant. No existing tests gamed. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — same `_finalize_sources` signature and 5-step pipeline, the generator-scoped `emitted_sources` set after the yield, `sources_to_persist = emitted_sources or None`, and the exact test matrix (new test_sources_interrupted.py + unit class in test_sources_event.py). No drift. |

---

PR #339, opened 2026-06-10

coleam00:
Fixes #277

**Source run-id:** `0882cf84-6016-4846-8c15-0dbe39255ef3`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
+      Use Read, Grep, and Glob to explore the repository. Look for:
+      - Files and modules involved in the issue
+      - Existing patterns and conventions to follow
+      - Tests that cover the affected code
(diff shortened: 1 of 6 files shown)


coleam00:

A strong, well-engineered fix. It correctly diagnoses that source finalization was coupled to the [DONE] branch and refactors the dedup/collapse/cap/refusal pipeline into a single pure helper invoked on every stream exit path, guarded by a run-once flag — so persisted sources can no longer diverge from the live view on interrupted or errored streams. Subtle-correctness care is notably good: dict-copy purity to avoid double-collapse, ordering the sources frame ahead of the error payload to beat the frontend's break-on-error parser, and a finally-block fallback for cancellation paths. Tests are meaningful and directly pin the issue's core invariant (live == persisted) plus the refusal edge case.

The main weakness is scope hygiene in the diff: four unrelated benchmark-cell YAML files (~600 lines) were swept into the PR by the workflow's `git add -A`, which is harness-induced noise rather than a code-reasoning defect but still inflates the changeset. The plan also asked the implementer to flag the protected-file edit for human review in the PR body, which didn't happen. Neither detracts from the soundness of the actual fix, which faithfully follows a detailed plan and addresses the reported bug at its root.

## Scores

|---|---|---|
| solves_issue | 9/10 | Directly fixes the live-vs-reload mismatch by routing every stream exit path (DONE, mid-stream error, stream-end, cancellation) through one finalization pipeline, and the regression test asserts emitted sources == persisted sources for an interrupted stream. Covers dedup, per-video collapse, and refusal-attached-chips exactly as the issue described. |
| root_cause | 9/10 | True root cause: the finalization logic (dedup -> is_cited -> collapse -> cap -> refusal-gate) was welded into the [DONE] branch, so interrupted streams persisted raw/intermediate state via a different path. Extracting it into a single pure _finalize_sources() applied on all exit paths with a sources_finalized flag guaranteeing exactly-once is the structurally correct fix, not a band-aid. |
| scope_discipline | 6/10 | The production change is surgical (one helper + call sites in messages.py, plus tests). However the diff also sweeps in four unrelated benchmark harness YAMLs (FF/FK/FO/OF, ~600 lines) via the workflow's `git add -A` — noise unrelated to issue #277. Harness-induced rather than a reasoning failure, but it still bloats the PR diff. |
| subtle_correctness | 9/10 | Thoughtful about foot-guns: purity via dict(tc) copies so re-invocation never sees half-collapsed state (segment_count double-count hazard), sources frame ordered before the error payload to survive the frontend's break-on-error parser, defensive stream-ended-without-DONE branch, and finally-block fallback for cancellation. Refusal text reconstructed from partial stream correctly handles interrupted refusals. Types and async generator semantics are clean. |
| code_quality | 9/10 | Production-grade: well-named pure helper with a clear docstring, idiomatic to the existing module, comments updated to describe the single-finalize invariant, and the route docstring step 7 updated. Matches the surrounding code's style and conventions. |
| test_discipline | 9/10 | Strong targeted tests: a TestFinalizeSources unit class (dedup order, collapse, cap, cited-passthrough, refusal, empty, purity/idempotence) plus a TestInterruptedStreamSources integration class that reproduces the exact bug and asserts live-emitted == persisted, no duplicate video/chunk ids, and refusal-persists-None. No existing tests gamed. |
| plan_impl_fidelity | 9/10 | The implementation executes the plan almost verbatim — same helper signature, same five call-site edits, same docstring update, and the exact test classes/scenarios the plan enumerated. Minor drift: the plan said to flag the protected-path (messages.py) edits for human review in the PR body, which the PR body omits. |

---

PR #340, opened 2026-06-10
[bench-FO] impl: address issue

coleam00:
Fixes #279

**Source run-id:** `fdf0a476-eb0d-4589-98cd-619c53979fa7`

### DIFF
--- a/app/backend/alembic/versions/0006_add_conversation_video_scope.py
+++ b/app/backend/alembic/versions/0006_add_conversation_video_scope.py
@@ -1,0 +1,34 @@
+"""Add per-conversation video scope to conversations.
+
+Schema change for issue #279 (scope a conversation to specific videos):
+- conversations: scoped_video_ids TEXT[] (nullable, no default)
+
+NULL (or an empty array, normalized to NULL at write time) means the
+conversation is unscoped and retrieval searches the whole library — today's
+behavior. A non-null array restricts retrieval and citations to those video
+ids. The column is nullable with no default so all existing rows upgrade
+in-place as unscoped.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE conversations ADD COLUMN IF NOT EXISTS scoped_video_ids TEXT[]")
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE conversations DROP COLUMN IF EXISTS scoped_video_ids")
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -253,6 +253,7 @@ async def keyword_search(
     top_k: int,
     language: str = "english",
     allowed_source_types: list[str] | None = None,
+    video_ids: list[str] | None = None,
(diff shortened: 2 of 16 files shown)


coleam00:

A strong, well-engineered PR that faithfully implements a thoughtfully scoped plan. The core strength is correct, layered enforcement: scope is filtered at the SQL WHERE clause on both the keyword and vector RRF legs and intersected into the transcript whitelist, so neither retrieval nor a full-transcript pull can escape the scope — and citations, derived from retrieved chunks, stay in-scope automatically. Edge cases are handled with care: empty-array normalization is consistent from API to DB, the scope filter composes with the member/source-type ACL via AND (never widening access), and a genuinely subtle fail-safe keeps the transcript tool guarded even when the library load fails. Tests are meaningful and cover the new behavior across every layer.

Weaknesses are minor. The PR necessarily edits three CLAUDE.md-protected files; this is inherent to the feature and was flagged for human review, but it means the change can't be fully factory-autonomous. The ChatArea scope button uses inline styles rather than Tailwind utilities (matching the adjacent Export button, but against the stated convention), and no automated frontend test was added for the picker (allowed by the plan's fallback). None of these affect correctness. Overall this is a high-fidelity, root-cause solution that cleanly solves issue #279.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully delivers the feature: a per-conversation scope persisted in the DB, SQL-level retrieval filtering on both RRF legs plus transcript-whitelist intersection so answers AND citations only come from picked videos, empty/null = whole library (today's behavior), and a frontend picker. Scope persists on the conversation row so it 'sticks' for every subsequent message. |
| root_cause | 9/10 | Enforcement is layered at the right place — the SQL WHERE clause on both keyword and vector legs (so RRF can't leak), plus intersecting the transcript whitelist with the scope set. Not a UI-only band-aid; the model literally cannot retrieve out-of-scope chunks. |
| scope_discipline | 8/10 | Surgical — every changed file is necessary and diffs are minimal (one endpoint, one read-and-thread block, two WHERE branches, one writer, one picker component). It does touch CLAUDE.md protected paths (conversations.py, messages.py, repository.py), but that is intrinsic to the feature and the plan explicitly flagged it for human review. |
| subtle_correctness | 9/10 | Thoughtful about failure modes: []->None normalization end-to-end so a cleared scope == never-scoped, both RRF legs filtered, ACL composition via AND (scope can never widen source-type access), and the standout — when the library whitelist load fails, scope falls back to the scope set itself rather than leaving the transcript tool unguarded. Unknown-id validation at write time avoids the silent matches-nothing trap. |
| code_quality | 9/10 | Production-grade: clear docstrings citing the issue, idiomatic asyncpg positional-param branches matching existing style, owner-scoped writer mirroring update_conversation_title, and a clean React modal consistent with CitationModal conventions. Inline styles on the ChatArea button are slightly off the Tailwind-only guideline but match the existing Export button's pattern. |
| test_discipline | 9/10 | Comprehensive targeted tests: endpoint happy path / []-normalization / unknown-id 422 / non-owner 404, message-flow threading + whitelist intersection, the whitelist-load-failure fallback, repository SQL clause presence/omission, retriever forwarding, and tool dispatch. Existing tests were updated only to add the new optional param to mock signatures — legitimate, not gaming. Frontend Vitest test was omitted, but the plan explicitly allowed that fallback. |
| plan_impl_fidelity | 10/10 | Near-exact execution of the plan: every listed file, the migration shape, the two-branch SQL approach, the writer signature, the messages.py read-and-intersect block (including the fallback-to-scope-set nuance), the endpoint with 422/404 semantics, and the frontend wrapper + picker all match item-for-item. Only the optional frontend test was skipped, which the plan permitted. |

---

PR #341, opened 2026-06-10
[bench-OO] feat(conversations): scope a conversation to specific videos

coleam00:
Fixes #279

**Source run-id:** `d27f5e77-41a5-46ab-8599-b0f18ebea372`

### DIFF
--- a/app/backend/alembic/versions/0006_add_scoped_video_ids_to_conversations.py
+++ b/app/backend/alembic/versions/0006_add_scoped_video_ids_to_conversations.py
@@ -1,0 +1,37 @@
+"""Add per-conversation video scope (issue #279).
+
+Adds a nullable ``scoped_video_ids TEXT[]`` column to ``conversations``. When a
+conversation is created with a non-empty list, every retrieval path for that
+conversation's messages filters chunks to those ``video_id``s. NULL/empty means
+"search the whole library" (the pre-#279 default), so existing rows upgrade
+in-place with no behavior change.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-05-31
+
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "conversations",
+        sa.Column("scoped_video_ids", postgresql.ARRAY(sa.Text()), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("conversations", "scoped_video_ids")
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -253,6 +253,7 @@ async def keyword_search(
     top_k: int,
     language: str = "english",
     allowed_source_types: list[str] | None = None,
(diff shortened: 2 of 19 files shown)


coleam00:

An excellent, high-fidelity implementation of issue #279. The change adds a nullable scoped_video_ids TEXT[] column and threads an additive video_id_filter through the entire retrieval chain (messages -> execute_tool -> executors -> retrieve_hybrid/keyword_search/vector_search_pg), enforcing scope at the SQL layer so both answers and citations are confined to the selected videos. It is careful about the details that matter: the unscoped path is kept byte-for-byte identical via conditional SQL, empty lists normalize to NULL, the filter composes with the existing member ACL rather than replacing it, and the transcript tool is guarded by intersecting the whitelist with the scoped set. Protected ownership files are touched only additively, preserving user_id checks. Tests are meaningful and span repository, tool, route, and frontend layers.

Weaknesses are minor: the banner/button use inline styles (consistent with surrounding code but less ideal than the Tailwind used elsewhere), and scope is creation-only with a read-only banner rather than any post-hoc view of which titles are in scope — both deliberate per the plan and the issue's immutability requirement. The implementation tracks the plan almost perfectly and even strengthens the transcript guard beyond what was planned. A clean, production-grade PR.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully delivers the issue: a per-conversation video scope set at creation, immutable thereafter, that filters all retrieval (hybrid/keyword/semantic) at the SQL layer so both answers and citations are confined to the chosen videos; empty/None searches the whole library. Includes a picker UI, scope banner, and transcript-tool guard. Minor gap only in that scope is creation-only with no in-UI confirmation of which titles are scoped, but that matches the issue's 'sticks' requirement. |
| root_cause | 9/10 | Structurally sound: enforcement is at the SQL WHERE clause (video_id = ANY(...)) on every search path plus a transcript whitelist intersection, threaded as an additive video_id_filter mirroring the existing allowed_source_types ACL pattern. Not a band-aid — the filter is applied where retrieval actually happens, so it can't be bypassed. |
| scope_discipline | 9/10 | Surgical and additive across exactly the files the change requires (migration, repository, retriever, tools, two routes, four frontend files). No drift into unrelated footguns, no refactors, ownership checks left verbatim. |
| subtle_correctness | 9/10 | Careful about subtle modes: conditional SQL/args so the unscoped path stays byte-for-byte identical (no extra param/clause), empty-list normalized to NULL, transcript whitelist intersected (not replaced) so stale ids drop out, scope composes with member ACL rather than replacing it, and the executor captures scope once per turn. No obvious async/type foot-guns. |
| code_quality | 9/10 | Readable, well-documented with issue-referencing comments, idiomatic to the codebase (keyword-only params, repository-only SQL, api.ts-only fetch). VideoScopePicker is clean Tailwind; the only nit is the ChatArea banner and Sidebar button use inline styles, though that matches the surrounding inline-style code in those files. |
| test_discipline | 9/10 | Strong targeted coverage at every layer: repository (round-trip, empty→NULL, scoped vs unscoped SQL), tools (filter forwarding + transcript rejection + unscoped None), route handler, and a frontend picker test asserting scoped vs unscoped createConversation calls. Existing-test edits only add the new param to mock signatures — a legitimate consequence of the signature change, not gaming. |
| plan_impl_fidelity | 10/10 | The diff executes the plan step-for-step: same migration, same additive video_id_filter threading, same conditional-SQL approach, same route/frontend changes, same four test files. It even improves on the plan's transcript step by intersecting the whitelist (& set(scoped)) instead of replacing it, while honoring every out-of-scope guardrail. |

---

PR #342, opened 2026-06-10

coleam00:
Fixes #279

**Source run-id:** `782a50c4-6e54-4e66-adfb-0ba5b1c1caac`

### DIFF
--- a/app/backend/alembic/versions/0006_conversation_scoped_videos.py
+++ b/app/backend/alembic/versions/0006_conversation_scoped_videos.py
@@ -1,0 +1,36 @@
+"""Add scoped_video_ids to conversations for per-conversation video scoping.
+
+Schema change for issue #279 (scope a conversation to specific videos):
+- conversations: scoped_video_ids TEXT[] — NULL means "unscoped, search all
+  videos" (the pre-#279 behaviour); a non-empty array is a hard retrieval
+  filter for the life of the conversation. No default and no index: the
+  column is only ever read per-conversation by primary key.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-06-10
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        ALTER TABLE conversations
+        ADD COLUMN IF NOT EXISTS scoped_video_ids TEXT[]
+        """
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE conversations DROP COLUMN IF EXISTS scoped_video_ids")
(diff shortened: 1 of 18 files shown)


coleam00:

An exemplary implementation. The PR threads a per-conversation video scope cleanly from a new nullable TEXT[] column through the repository, hybrid retriever, tool executors, and message route, with write-once immutability enforced at the database layer and the transcript whitelist intersected with the scope so neither answers nor citations can escape the selected videos. It solves the issue at the root rather than papering over it, and is meticulous about subtle edges — empty-array vs NULL collapse, dedup/normalization, deleted-video pruning, and a race-safe IS NULL guard. Test coverage is broad and meaningful at every layer, and fidelity to the upfront plan is near-perfect.

Weaknesses are minor: ChatArea introduces a sizeable inline-style block where the codebase nominally prefers Tailwind (mitigated by that file's existing convention), and the write-once design means a user cannot re-scope or scope a conversation that already has messages — a deliberate, defensible reading of the issue's 'stick for the rest of the conversation' requirement but a slight UX constraint. Overall a production-grade change.

## Scores

|---|---|---|
| solves_issue | 9/10 | Fully delivers per-conversation video scoping: scope at creation or via a write-once endpoint, both retrieval legs filtered, transcript whitelist narrowed to scope, and a frontend picker. Since citations are built from the returned chunks, scoping retrieval also scopes citations; default-unscoped preserves search-everything. Minor gap: scope can only be set on an empty conversation (write-once), which matches the issue's stickiness intent but slightly narrows 'pick a subset' UX. |
| root_cause | 9/10 | Structurally sound — a nullable scoped_video_ids column is the single source of truth, filtering happens at the SQL layer on both keyword and vector legs, and write-once immutability is enforced by the `scoped_video_ids IS NULL` UPDATE guard at the DB level rather than in app logic. Not a band-aid. |
| scope_discipline | 9/10 | 18 files but every one is justified and mirrors the plan exactly; protected-path files (messages.py, conversations.py, repository.py) are touched only in retrieval plumbing with user_id scoping and rate-limit call sites untouched. No drive-by refactors. |
| subtle_correctness | 9/10 | Careful about the footguns: `[] or None` collapse so asyncpg's empty array never becomes a filter, belt-and-suspenders normalization at multiple layers, whitelist = scope ∩ library so a deleted scoped id is dropped, and the DB-level race-safe write-once guard. Validation rejects unknown ids before any write. |
| code_quality | 8/10 | Readable, well-documented with issue references, idiomatic to the codebase, static SQL variants instead of string interpolation. Slight ding: ChatArea adds a large block of inline styles for the chip/button (CLAUDE.md prefers Tailwind), though it matches that file's existing inline-style convention; VideoScopePicker correctly uses Tailwind. |
| test_discipline | 9/10 | Comprehensive, targeted tests across every layer: repository SQL shape (scoped + regression unscoped), retriever forwarding, tools forwarding, route-level scope lifecycle (201/409/404/422), message-route threading and whitelist intersection, plus frontend picker and api-wrapper tests. Existing tests updated only for the new signature param — legitimate, not gamed. |
| plan_impl_fidelity | 10/10 | The diff executes the plan point-for-point — migration 0006, repository params and set_conversation_scope write-once UPDATE, retriever/tools threading, the conversations scope endpoint with 422/409/404 disambiguation, messages.py whitelist intersection, and every named frontend file and test. Out-of-scope items (catalog filtering, Sidebar, retriever.py) were correctly left untouched. |

---

PR #343, opened 2026-06-10
[bench-FK] review: address self-review findings

coleam00:
Fixes #279

**Source run-id:** `7b00b255-0e9c-4cef-a06c-efc2c76d918a`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
+      Use Read, Grep, and Glob to explore the repository. Look for:
+      - Files and modules involved in the issue
+      - Existing patterns and conventions to follow
+      - Tests that cover the affected code
+      - Constraints in CLAUDE.md, MISSION.md, or FACTORY_RULES.md
+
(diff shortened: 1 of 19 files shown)


coleam00:

A strong, well-executed PR that fully delivers the conversation-scoping feature. The fix is structurally sound: it filters at the SQL layer by extending the existing allowed_source_types pattern with allowed_video_ids, guarantees citations come only from scoped videos for free, enforces set-once persistence atomically via a `video_scope IS NULL` UPDATE predicate, and closes the transcript-tool bypass the exploration flagged. Edge cases (empty array vs NULL, ownership 404, set-once races, unknown-id 422 validation) are handled thoughtfully, and test coverage is broad and targeted across the retriever, tool, dispatcher, and route layers. Fidelity to the plan is near-total.

Weaknesses are minor. The diff includes four unrelated benchmark-harness YAML files that inflate it and hurt pure scope discipline, though these are clearly orchestration scaffolding rather than the implementer overreaching into the codebase. Code quality is high overall; the only nits are verbose inline styles in ChatArea (consistent with that file but against the stated Tailwind convention) and a redundant emoji-plus-icon button. No frontend unit tests, but that matches the repo's missing vitest setup and was acknowledged in the plan.

## Scores

|---|---|---|
| solves_issue | 9/10 | Implements the feature end-to-end: scope set at creation or once via PATCH, persisted in a nullable video_scope column, threaded into every retrieval path and the transcript tool, with NULL preserving whole-library behavior. Because filtering happens at the SQL layer, both answers and citations are restricted to the chosen videos exactly as the issue asks. |
| root_cause | 9/10 | Structurally correct: reuses the existing allowed_source_types ANY($n::text[]) filter pattern at the SQL boundary rather than post-filtering in Python, and closes the transcript-tool bypass with a scope guard. Set-once is enforced atomically via the `video_scope IS NULL` UPDATE predicate, not just an app-level check. |
| scope_discipline | 6/10 | The feature code itself is surgical and matches the planned file list. However, the PR diff also carries four unrelated benchmark harness YAMLs (benchmark-FF/FK/FO/OF, ~600 lines) that have nothing to do with issue #279; they are orchestration scaffolding rather than feature creep, but they still bloat the diff. |
| subtle_correctness | 8/10 | Handles the tricky cases well: empty-array vs NULL normalized in both the validator and messages.py (`list(scope) if scope else None`), asyncpg native list↔TEXT[] mapping, set-once race resolved to 409, ownership 404 preserved, and the transcript guard layered after the whitelist check. Whitespace-only ids silently collapse to unscoped, a minor benign edge. |
| code_quality | 8/10 | Backend is idiomatic to the repo — parameterized $n SQL, updated docstrings, additive changes to protected paths. The new VideoScopePicker correctly uses Tailwind-only styling; the inline-style ScopeChip/Focus button in ChatArea deviate from the stated Tailwind rule but match that file's existing inline-style convention. The 🎯 emoji beside an SVG icon is mildly redundant. |
| test_discipline | 9/10 | Adds a focused 752-line test file covering retriever forwarding, executor/dispatcher threading, transcript guard (in/out/none), route validation (422 unknown/empty, 200, PATCH 200/409/404), and message-route scope threading for both set and null scope. Existing-test edits only add the new optional param to mock signatures — legitimate, not gamed. No frontend tests, consistent with the absent vitest setup. |
| plan_impl_fidelity | 9/10 | The implementation tracks the plan almost item-for-item: migration 0006, repository create/set/search filters, retriever_hybrid threading, tools threading + transcript guard, ConversationCreate validation + new PATCH /scope endpoint, messages.py 4-line thread, api.ts wrappers, VideoScopePicker, ChatArea affordance + chip, useMessages field. Only cosmetic drift (chip placement centered vs 'header area'). |

---

PR #344, opened 2026-06-10

coleam00:
Fixes #279

**Source run-id:** `19e40c0e-cdb5-49b4-a679-fd9ff73c257b`

### DIFF
--- a/app/backend/alembic/versions/0006_add_conversation_video_scope.py
+++ b/app/backend/alembic/versions/0006_add_conversation_video_scope.py
@@ -1,0 +1,37 @@
+"""Add per-conversation video scope (issue #279).
+
+Adds `scoped_video_ids TEXT[]` to `conversations`. NULL (the default for all
+existing and new rows) means the conversation is unscoped and retrieval
+searches the whole library — the pre-#279 behavior. A non-empty array
+restricts retrieval and citations to those videos only.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-06-10
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    # Nullable, no default: NULL = unscoped. An empty array is never stored
+    # (the API normalizes [] to NULL) to avoid the "scope to nothing" ambiguity.
+    op.execute(
+        """
+        ALTER TABLE conversations
+        ADD COLUMN IF NOT EXISTS scoped_video_ids TEXT[]
+        """
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE conversations DROP COLUMN IF EXISTS scoped_video_ids")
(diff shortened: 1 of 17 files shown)


coleam00:

A strong, faithful implementation of per-conversation video scoping. The fix is structurally correct: it adds a nullable scoped_video_ids column, filters at the SQL layer additively with the existing source_type ACL, and adds defense-in-depth so out-of-scope transcripts can't be pulled directly. Edge-case handling is thoughtful — a single NULL representation for 'unscoped', empty-list-to-None normalization, dedup with order preservation, and proper 400/404 responses — and the scope correctly 'sticks' by being re-read from the row on every turn. Test coverage is broad and meaningful across backend tool, route, and message layers plus frontend api and component tests.

Weaknesses are minor. The ChatArea diff is larger than strictly necessary because the export button was restructured into a flex container, and the inline-style blocks are duplicated. The PR touches several CLAUDE.md-protected paths (messages.py, repository.py, conversations.py); this is inherent to the issue and the plan flagged it, but the PR body never includes the promised in-scope justification note. There is also no live-DB test proving the row-level filter actually excludes out-of-scope chunks (deferred by the plan to a manual smoke note). None of these undermine correctness, and the implementation tracks the plan almost line-for-line.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers the feature end-to-end: a per-conversation scoped_video_ids column, a PATCH endpoint, SQL-layer filtering on all three search paths plus transcript enforcement, and a frontend picker/indicator. Scope persists on the row and is re-read each turn so it 'sticks'; unscoped (NULL) preserves whole-library search exactly as before. |
| root_cause | 9/10 | Structurally sound rather than a band-aid: filtering happens at the SQL WHERE clause mirroring the existing source_type ACL, applied additively (ACL still enforced), with defense-in-depth blocking out-of-scope transcripts. NULL = unscoped is a clean single-representation invariant enforced at both write and API layers. |
| scope_discipline | 8/10 | Touches only files the feature requires and mostly threads one parameter through. The ChatArea export-button restructure into a flex container inflates the diff somewhat but is necessary to add the adjacent scope button. Touches protected paths (messages.py, repository.py, conversations.py) which the issue inherently requires, though the PR body omits the planned acknowledgement note. |
| subtle_correctness | 9/10 | Careful about the foot-guns: empty list normalized to None at binding (video_ids if video_ids else None), single NULL representation for unscoped, dedup preserving order, all-unknown rejected with 400, cross-user 404, asyncpg list->TEXT[] mapping, and scope re-read every turn for no per-message state. No obvious async/type bugs. |
| code_quality | 8/10 | Readable, well-documented with issue references, idiomatic to the codebase (inline styles match existing ChatArea conventions, Tailwind in the new modal, named exports). Minor duplication of inline button-style blocks in ChatArea; otherwise production-grade. |
| test_discipline | 9/10 | Extensive, targeted tests across every layer: tool-threading forwarding, transcript in/out-of-scope, PATCH validation (subset, clear variants, all-unknown 400, cross-user 404), message-turn scope propagation, and frontend api + picker tests. Existing tests updated only to add the new kwarg signature, not gamed. Only gap is no live-DB row-level filter test, which the plan explicitly deferred. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — every numbered backend and frontend item (migration 0006, repository filters + update_conversation_scope, retriever forwarding, tools threading + transcript guard, messages read/forward, PATCH endpoint, api.ts wrapper, useMessages setConversation, VideoScopePicker, ChatArea button) and the full test matrix are present as specified. |

---

PR #347, opened 2026-06-10

coleam00:
Fixes #280

**Source run-id:** `fb36f787-50d0-40b1-9933-4080cc90cef1`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -603,6 +603,32 @@ async def list_messages(conversation_id: str, user_id: str) -> list[dict]:
     return results
 
 
+async def delete_last_assistant_message(conversation_id: str, user_id: str) -> str | None:
+    """Delete the most recent assistant message in a conversation (owner-scoped).
+
+    Returns the deleted message id, or None if there is no assistant message
+    or the conversation does not belong to the user.
+    """
+    async with _acquire() as conn:
+        row = await conn.fetchrow(
+            """
+            DELETE FROM messages
+            WHERE id = (
+                SELECT m.id FROM messages m
+                JOIN conversations c ON c.id = m.conversation_id
+                WHERE m.conversation_id = $1 AND c.user_id = $2
+                  AND m.role = 'assistant'
+                ORDER BY m.created_at DESC
+                LIMIT 1
+            )
+            RETURNING id
+            """,
+            conversation_id,
+            user_id,
+        )
+    return row["id"] if row else None
+
+
 # ---------------------------------------------------------------------------
 # Channel sync runs
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


coleam00:

A high-quality, near-textbook implementation of issue #280. The feature is fully solved: regenerate replaces the latest answer in place, re-streams with its own citations through the existing SSE machinery, and is counted fairly against the 25 msg/24h cap. The standout strength is the security/correctness discipline — the rate-limit check_and_record runs before any mutation so an over-cap user can neither bypass the counter nor lose their old answer, and the DELETE is owner-scoped via a JOIN on conversations. The refactor into a shared _stream_assistant_response helper avoids duplicating battle-tested streaming logic, and the optimistic-remove-then-restore frontend flow correctly handles rate-limit, abort, and generic errors. Test coverage is excellent across backend route, repository SQL shape, and three frontend layers.

Weaknesses are minor: the regenerate route reads list_messages twice (once for preconditions, once inside the helper), a small inefficiency; and the new Regenerate button leans on inline styles and JS hover handlers rather than the Tailwind-only convention in CLAUDE.md, though it matches the styling of pre-existing inline buttons. The implementation tracks the plan almost line-for-line, indicating the Opus plan and Fable implementation were well-coordinated. Overall a strong, mergeable PR.

## Scores

|---|---|---|
| solves_issue | 10/10 | Delivers exactly what #280 asked: a Regenerate button on the latest assistant answer that deletes the stale reply, streams a fresh one with its own sources via the identical SSE path, and counts against the daily cap — including a no-quota-bypass guarantee. |
| root_cause | 9/10 | Structurally sound: a dedicated /regenerate endpoint plus extraction of the shared load→stream→persist block into _stream_assistant_response, reusing the same ownership→rate-limit→mutate invariant ordering rather than duplicating or hacking the send path. |
| scope_discipline | 9/10 | Surgical — only the 6 files the feature needs (repository, route, two hooks/components, plus tests). No drift into unrelated areas; repository change is purely additive and respects the SQL-in-repository.py rule. |
| subtle_correctness | 9/10 | Careful about the failure modes: rate-limit recorded before any delete (no bypass), 409 pre-checks avoid burning quota on no-ops, owner-scoped DELETE via JOIN, and optimistic removal restored on RateLimitError/AbortError/generic error. Minor: list_messages is read twice (route + helper). |
| code_quality | 8/10 | Readable, well-commented, idiomatic to the codebase with clear invariant docs. Slight deviation: the Regenerate button uses heavy inline styles and onMouseEnter hover handlers rather than Tailwind utility classes per CLAUDE.md (though it mirrors existing inline buttons). |
| test_discipline | 10/10 | Thorough, targeted tests at every layer: backend route (happy path, rate-limit-counts, over-cap-no-delete, 409 preconditions, cross-user 404), repository SQL-shape test, and frontend hook/Message/ChatArea behavior tests covering optimistic-remove/restore. Nothing gamed. |
| plan_impl_fidelity | 10/10 | The diff executes the source plan almost verbatim — same endpoint, same _stream_assistant_response extraction, the exact delete_last_assistant_message SQL, the regenerate option on useStreamingResponse, and every planned test scenario is present. |

---

PR #348, opened 2026-06-10

coleam00:
Fixes #280

**Source run-id:** `3406e69e-5af7-40f2-884c-58cc7fdb81ae`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -603,6 +603,36 @@ async def list_messages(conversation_id: str, user_id: str) -> list[dict]:
     return results
 
 
+async def delete_last_assistant_message(conversation_id: str, user_id: str) -> str | None:
+    """Delete the conversation's most recent message iff its role is 'assistant'.
+
+    Owner-scoped (JOIN on conversations.user_id) and atomic: the role check on
+    the DELETE itself (not just the subquery) makes concurrent regenerates safe —
+    a second request's subquery resolves to the now-last user message, the role
+    predicate fails, zero rows are deleted, and None is returned.
+
+    Returns the deleted message id, or None when nothing was deleted.
+    """
+    async with _acquire() as conn:
+        row = await conn.fetchrow(
+            """
+            DELETE FROM messages
+            WHERE id = (
+                SELECT m.id FROM messages m
+                JOIN conversations c ON c.id = m.conversation_id
+                WHERE m.conversation_id = $1 AND c.user_id = $2
+                ORDER BY m.created_at DESC, m.id DESC
+                LIMIT 1
+            )
+            AND role = 'assistant'
+            RETURNING id
+            """,
+            conversation_id,
+            user_id,
+        )
+    return row["id"] if row else None
+
+
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


coleam00:

A strong, well-engineered PR that fully solves issue #280. The regenerate endpoint reuses a cleanly-extracted shared streaming helper rather than duplicating ~140 lines, deletes the trailing assistant message via an atomic, owner-scoped DELETE whose role predicate makes concurrent regenerates race-safe, and charges the rate limit before any mutation so a 429 provably leaves the conversation untouched — which the frontend leverages to restore the original answer locally. Test discipline is excellent across both backend and frontend, covering ownership, regenerability, quota, the lost-race, and refusal suppression. Plan-implementation fidelity is near-perfect.

Weaknesses are minor. The RegenerateButton is styled with a large static inline style object instead of Tailwind utilities, a small deviation from the codebase's stated convention (mitigated by intentionally matching the existing sources-toggle button). The PR also edits two protected paths (messages.py, repository.py), which the plan acknowledges and confines to additions plus pure code motion — appropriate here since the feature genuinely requires them and the author is human, but worth flagging. Overall a high-quality cell.

## Scores

|---|---|---|
| solves_issue | 10/10 | Delivers exactly what #280 asked: a regenerate button on the last assistant message, the old answer is replaced by a fresh stream with its own sources, identical SSE experience, and it charges the same 25/24h quota so users can't sneak past their limit. |
| root_cause | 9/10 | Structurally sound: a dedicated endpoint reuses a properly-extracted shared streaming helper (no copy-paste), an atomic owner-scoped DELETE handles regeneration, and the rate-limit check runs before any mutation. Not a band-aid. |
| scope_discipline | 9/10 | Surgical — only the 9 files the feature needs, additions-only in the two protected files, no unrelated footgun cleanup. The streaming-core extraction is pure code motion. |
| subtle_correctness | 9/10 | Genuinely thoughtful about failure modes: AND role='assistant' on the DELETE makes concurrent regenerates race-safe (returns None→409), 429 fires before the delete so the frontend restore is server-consistent, refusal suppression and asyncio.shield persist are preserved. Validation rejects empty/user-last/no-prior-user before charging quota. |
| code_quality | 8/10 | Readable, well-commented, idiomatic with strong rationale in docstrings. Minor ding: RegenerateButton uses a large static inline style object rather than Tailwind utilities (CLAUDE.md prefers Tailwind), though it intentionally mirrors the existing sources-toggle button's styling. |
| test_discipline | 10/10 | Comprehensive, targeted tests: backend covers 404, three 409 variants (no quota charge), 429 (conversation intact), happy path (quota once, history excludes deleted answer, sources persisted), lost-race 409, and refusal suppression; frontend covers the hook, Message button gating, and ChatArea remove/restore/429 flows. Nothing gamed. |
| plan_impl_fidelity | 10/10 | The diff faithfully executes the plan end-to-end: delete_last_assistant_message with the exact SQL, _stream_assistant_response extraction, the ordered regenerate guard flow, runStream/startRegenerate refactor, the Message prop+button, ChatArea handleRegenerate with restore-on-429, and the full test matrix — all as specified, no drift. |

---

PR #349, opened 2026-06-10

coleam00:
Fixes #280

**Source run-id:** `c6feb84b-f8c0-4630-940d-1bd17b2e6f62`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -577,6 +577,28 @@ async def create_message(
     }
 
 
+async def delete_message(message_id: str, conversation_id: str, user_id: str) -> bool:
+    """Delete a message. Returns False if the message or owning conversation
+    doesn't match (atomic ownership check, same pattern as create_message)."""
+    async with _acquire() as conn:
+        result = await conn.execute(
+            """
+            DELETE FROM messages
+            WHERE id = $1 AND conversation_id = $2
+              AND EXISTS (
+                  SELECT 1 FROM conversations WHERE id = $2 AND user_id = $3
+              )
+            """,
+            message_id,
+            conversation_id,
+            user_id,
+        )
+        if result == "DELETE 0":
+            return False
+    await touch_conversation(conversation_id, user_id)
+    return True
+
+
 async def list_messages(conversation_id: str, user_id: str) -> list[dict]:
     """Return messages only if the conversation belongs to the given user."""
     async with _acquire() as conn:
--- a/app/backend/routes/messages.py
+++ b/app/backend/routes/messages.py
@@ -54,74 +54,22 @@ class MessageCreate(BaseModel):
 
 
 # ---------------------------------------------------------------------------
-# POST /api/conversations/{conv_id}/messages
+# Shared streaming helper
 # ---------------------------------------------------------------------------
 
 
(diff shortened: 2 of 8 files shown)


coleam00:

An excellent, production-grade PR that cleanly solves issue #280. The core design — a dedicated regenerate endpoint that reuses the existing SSE/tool/persistence path via an extracted `_stream_assistant_response` helper — avoids duplicating ~150 lines while keeping the original `create_message` semantics byte-identical. The security and fairness invariants are handled with care: ownership checks return 404 without leaking existence, validation returns 409 for invalid histories, and the rate-limit gate runs before the destructive delete so a capped user never loses their existing answer. The frontend mirrors this rigor with optimistic removal and full restore-on-error paths (abort, rate-limit, generic), and the streaming UX flows through the established `useStreamingResponse` hook per CLAUDE.md conventions. Testing is thorough and meaningful on both tiers, including the structural user_id scoping invariant for the new repo function. Weaknesses are minor: the Message.tsx button leans on inline styles (consistent with the existing sources-toggle, but not Tailwind-idiomatic per CLAUDE.md), and the new endpoint adds surface to the protected messages.py/repository.py paths — additive and well-justified, but worth a human eye given the protected-path rules. Overall a near-textbook execution of a clear plan.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers exactly what the issue asked: a regenerate button on the latest assistant message that streams a fresh answer (with its own sources) to replace the old one, and counts as one normal send against the 25/24h cap. All 'done looks like' criteria are met end-to-end (route + hook + UI). |
| root_cause | 9/10 | Structurally sound: a dedicated ownership-scoped endpoint that reuses the existing streaming machinery via an extracted shared helper, plus a properly atomic `delete_message`. No symptom-papering — the rate-limit gate runs before the delete so a capped user keeps their answer, which is the correct invariant. |
| scope_discipline | 9/10 | Surgical and additive: touches only the 8 files the work required, with no drift into unrelated code. The `create_message` body is moved verbatim into the helper rather than rewritten, keeping the existing route's semantics untouched. |
| subtle_correctness | 9/10 | Thoughtful about failure modes: rate-limit-before-delete ordering, 404 vs 409 distinction (no existence leak), 409 when last message isn't an assistant preceded by a user, atomic EXISTS ownership guard, and frontend optimistic-restore on AbortError/RateLimitError/generic failure. No obvious foot-guns introduced. |
| code_quality | 9/10 | Reads like the surrounding code — matching comment density, docstrings, async patterns, and the existing inline-style button idiom in Message.tsx. Well-named helper and clear numbered-step comments mirror the original handler. |
| test_discipline | 9/10 | Comprehensive, targeted tests on both sides: backend covers 404/409 validation, happy-path delete+stream+persist, 429-leaves-answer-intact, quota consumption, and LLM history shape; frontend covers button placement, optimistic removal, replacement, and failure/429 restore. The scoping test also extends the structural user_id invariant to the new function. |
| plan_impl_fidelity | 10/10 | The implementation executes the plan almost verbatim — same endpoint, same helper-extraction strategy, same delete_message signature/SQL, same rate-limit ordering, same frontend wiring and option flag, and the exact test matrix described. No meaningful divergence. |

---

PR #352, opened 2026-06-10

coleam00:
Fixes #294

**Source run-id:** `3789f6d0-b2a9-411b-b2fb-19e63cdd4152`

### DIFF
--- a/app/backend/alembic/versions/0006_add_sources_gin_index.py
+++ b/app/backend/alembic/versions/0006_add_sources_gin_index.py
@@ -1,0 +1,33 @@
+"""Add a GIN index on messages.sources for the video filter (issue #294).
+
+The conversation list can now be filtered by video. The filter runs a JSONB
+containment subquery (`m.sources @> '[{"video_id": ...}]'`) over messages;
+without an index that is a full-table scan. `jsonb_path_ops` is the optimal
+operator class for `@>` containment-only lookups.
+
+Revision ID: 0006
+Revises: 0005
+Create Date: 2026-06-10
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0006"
+down_revision: str | None = "0005"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "CREATE INDEX IF NOT EXISTS idx_messages_sources_gin "
+        "ON messages USING gin (sources jsonb_path_ops)"
+    )
+
+
+def downgrade() -> None:
+    op.execute("DROP INDEX IF EXISTS idx_messages_sources_gin")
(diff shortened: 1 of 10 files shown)


coleam00:

A clean, well-scoped PR that fully solves issue #294 by relocating conversation filtering server-side so text, date-range, and video filters compose in one indexed SQL query while preserving ownership scoping, the preview subselect, and updated_at DESC ordering. The implementation is an almost verbatim execution of the Opus plan, with a sensible GIN index for the JSONB video lookup and disciplined, non-gamed tests across both tiers (SQL-construction capture tests, route pass-through, and frontend hook/api coverage), with realistic DB-dependent behavioral tests gated behind the repo's existing skip marker.

Weaknesses are minor and mostly cosmetic: the 'Last month' preset is a rolling 30-day window rather than a true calendar month, preset dates are computed in the browser's local timezone then compared against timestamptz on the server (a small boundary risk), unescaped ILIKE wildcards in the q parameter, and the JSON.parse(filtersKey) round-trip is a slightly awkward way to stabilize the callback identity. None of these undermine correctness for the stated use case, and the change respects every protected-path and out-of-scope constraint. Strong overall.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers exactly what the issue asked: date-range filtering (presets + custom), video filtering, both combining with text search via AND, and results still ordered updated_at DESC. Minor: 'Last month' preset maps to a rolling 30 days rather than a calendar month. |
| root_cause | 9/10 | Correctly moves filtering server-side into a single composable SQL query instead of patching the client-side title-only filter, and adds a GIN index for the JSONB video-containment subquery. This is the structurally right place for the fix. |
| scope_discipline | 9/10 | Surgical: exactly the six files the plan named plus the two consequential test files, no unrelated churn. Touches the protected routes/repository.py but strictly additively, leaving the user_id ownership guard byte-for-byte intact. |
| subtle_correctness | 8/10 | Thoughtful: stale-fetch guard preserved, filtersKey stringification prevents redundant refetches, inclusive-day date semantics documented, ownership guard always $1. Minor unaddressed edges: ILIKE treats user '%'/'_' as wildcards, and preset dates are computed in local TZ but compared against timestamptz server-side. |
| code_quality | 8/10 | Readable, well-documented (clear docstrings citing issue #294), idiomatic to the codebase's inline-style sidebar pattern and repository SQL conventions. The JSON.parse(filtersKey) round-trip in the load callback is slightly hacky, and the inline-style blocks are verbose, but consistent with existing idiom. |
| test_discipline | 9/10 | Strong: SQL-construction tests capture the actual query/args to verify additive AND composition and positional params without a DB, route pass-through is unit-tested, frontend hook/api tests are rewritten meaningfully, and behavioral DB tests are gated behind the same skip the repo already uses—no gaming. |
| plan_impl_fidelity | 10/10 | Near-perfect execution of the plan point-for-point: migration 0006 with jsonb_path_ops, list_conversations_filtered with the exact signature and clause semantics, route params, ConversationFilters in api.ts, useConversations refactor, Sidebar presets/dropdown/empty-state, and the full test matrix all delivered as specified. |

---

PR #354, opened 2026-06-10

coleam00:
Fixes #294

**Source run-id:** `10be0f82-aa61-4186-8e6f-2e32ec862b8f`

### DIFF
--- a/app/backend/db/repository.py
+++ b/app/backend/db/repository.py
@@ -745,3 +745,31 @@ async def list_sync_videos_for_run(sync_run_id: str) -> list[dict]:
             sync_run_id,
         )
     return [dict(r) for r in rows]
+
+
+# ---------------------------------------------------------------------------
+# Conversation video references (for sidebar filtering)
+# ---------------------------------------------------------------------------
+
+
+async def list_conversation_video_refs(user_id: str) -> list[dict]:
+    """Distinct (conversation_id, video_id) pairs cited in a user's messages.
+
+    Walks messages.sources JSONB (array of citation objects, each with a
+    video_id) joined through conversations for owner scoping.
+    """
+    async with _acquire() as conn:
+        rows = await conn.fetch(
+            """
+            SELECT DISTINCT m.conversation_id, src->>'video_id' AS video_id
+            FROM messages m
+            JOIN conversations c ON c.id = m.conversation_id
+            CROSS JOIN LATERAL jsonb_array_elements(m.sources) AS src
+            WHERE c.user_id = $1
+              AND m.sources IS NOT NULL
+              AND jsonb_typeof(m.sources) = 'array'
+              AND src->>'video_id' IS NOT NULL
+            """,
+            user_id,
+        )
+    return [dict(r) for r in rows]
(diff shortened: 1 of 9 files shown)


coleam00:

A strong, well-engineered PR that fully solves issue #294. The standout is its handling of the one hard part — 'which video was a conversation about' — via a clean additive read-only endpoint that queries messages.sources JSONB, owner-scoped through the conversations join, without touching any protected auth/scoping code or requiring a migration. Date filtering correctly reuses updated_at and is timezone-aware (local-day parsing avoids the classic UTC-midnight boundary bug), and all three filters compose while preserving the server's newest-first order. Test coverage is comprehensive and honest at both the hook and route layers. Plan fidelity is excellent; the implementation reads like a faithful execution of a careful plan.

Weaknesses are minor: the video filter fails open (renders all conversations) when the refs fetch errors rather than surfacing the empty/failed state cleanly, the refetch path duplicates the ref-map construction instead of factoring it out, and the empty-state copy was flattened to a generic message, dropping the planned per-query wording. None of these affect correctness of the core feature. Overall a high-quality cell.

## Scores

|---|---|---|
| solves_issue | 9/10 | Delivers exactly what #294 asked: date-range filtering (7d/30d presets + custom from/to), per-video filtering, combinable with the existing text search, and newest-first ordering is preserved because filters are pure array passes over the server-ordered list. Loses a point only because the video dropdown silently fails open (shows all conversations) if the refs fetch errors. |
| root_cause | 9/10 | Structurally sound. Date filtering reuses the already-present updated_at with no backend cost; the one genuinely missing relationship (which videos a conversation cites) is exposed via a clean, additive, owner-scoped GET /conversation-videos endpoint backed by a parameterized JSONB query — not a denormalization hack or a migration. |
| scope_discipline | 9/10 | Surgical: 9 files, all directly required. Deliberately routes around the protected conversations.py/repository.py scoping logic by appending a new function and a new route file, and avoids the /conversations/{id} route-capture trap. No unrelated footgun cleanup. |
| subtle_correctness | 8/10 | Thoughtful about the key foot-gun: parses YYYY-MM-DD into local-time Date (not UTC new Date(str)) so day boundaries don't shift across timezones, with inclusive start/end-of-day. Lazy refs fetch with a cancelled-guard and stale fetchId guard. Minor: video filter fails open on refs error, and refetch duplicates the map-building logic. |
| code_quality | 8/10 | Readable, well-named, idiomatic — matches the file's existing inline-style approach and the title-search filtering pattern; backend mirrors conversations.py. Docstrings are clear. Slight dings for verbose inline-style blocks and the duplicated ref-map construction in load vs refetch. |
| test_discipline | 9/10 | Meaningful, targeted tests at both layers: hook tests cover from-only/to-only/both inclusive boundaries, video membership, lazy-fetch-not-called, refs error path, combined intersection, and order preservation; backend covers 200 shape + 401; Sidebar covers preset/video/custom-reveal/empty-state. Existing tests updated for the new signature rather than gamed. |
| plan_impl_fidelity | 9/10 | Near-verbatim execution of the plan — same endpoint, same SQL, same filters-object hook signature, same file set, same test matrix. Only drift: the empty state was simplified to a single 'No conversations match your filters' for all active filters, dropping the plan's intent to keep the specific 'No matches for "query"' copy for text-only search. |

---

PR #355, opened 2026-06-10
[bench-OO] fix(rag): send Anthropic duration string for extended-tier cache ttl

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `88542df6-053f-4d03-8b50-7cbf5626b916`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
+      ## Instructions
+
+      Use Read, Grep, and Glob to explore the repository. Look for:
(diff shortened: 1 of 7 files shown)


coleam00:

A near-textbook fix for issue #246. The implementer correctly identified that Anthropic's cache_control.ttl is a duration-string enum ("5m"|"1h"), not an integer second count, and avoided the tempting-but-wrong stringify ("3600") trap. Crucially it treated the root cause structurally: the misleading CATALOG_CACHE_TTL_SECONDS config constant (and its incorrect 'requires an integer' comment) was removed rather than left to encode the wrong mental model, and the dead import was cleaned up to avoid a lint failure. Tests were updated to assert the real contract, with the previously-misnamed test rewritten honestly and a standard-tier guard added. Plan-to-implementation fidelity is essentially perfect.

The one real weakness is scope: the PR carries four unrelated benchmark workflow YAML files (~616 lines) that have nothing to do with the cache-ttl fix. These are almost certainly benchmark-harness artifacts pulled in from the worktree rather than the implementer's intentional scope creep, and they don't affect application behavior, but they do bloat the diff and would warrant being split out in a real review. Net: an excellent, root-cause fix marred only by harness-driven diff pollution.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control.ttl="1h" and standard tier still emits plain {"type": "ephemeral"} with no ttl. Catalog block text and inclusion logic untouched — every acceptance criterion is met exactly. |
| root_cause | 10/10 | Fixes the true root cause: it recognizes the config's "TTL in seconds" abstraction is itself wrong and removes CATALOG_CACHE_TTL_SECONDS entirely, mapping the tier to Anthropic's enum rather than stringifying or otherwise papering over the integer. |
| scope_discipline | 6/10 | The code fix is surgical (catalog.py, config.py, test_catalog.py), but the PR also bundles four unrelated ~155-line benchmark YAML files (FF/FK/FO/OF) — ~616 lines of harness infra with no bearing on issue #246. Likely swept in from the worktree, but it pollutes the diff. |
| subtle_correctness | 10/10 | Avoids the tempting wrong fix (stringifying to "3600"), correctly removes the now-dead import to prevent a ruff unused-import error, and leaves standard tier's ttl-less path intact. No foot-guns introduced. |
| code_quality | 9/10 | Idiomatic and minimal; the replacement comment ('Anthropic cache_control.ttl enum: "5m" | "1h"') correctly documents the API contract that the old comment got wrong. Clean removal of the dead constant. |
| test_discipline | 9/10 | Both affected tests are updated to assert the correct new contract (str "1h"), the misnamed integer-asserting test is renamed and rewritten rather than gamed, and a standard-tier no-ttl assertion is added as a guard. Tests verify real behavior, not just made-to-pass. |
| plan_impl_fidelity | 10/10 | The implementation faithfully executes every item of source-plan.md — import removal, "1h" replacement with the exact documented comment, config constant+comment removal, and the two test updates including the optional standard-tier no-ttl assertion. No drift. |

---

PR #356, opened 2026-06-10
[bench-FO] impl: address issue

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `48042a22-61a3-4c50-9eb9-d050fde267b5`

### DIFF
--- a/app/backend/config.py
+++ b/app/backend/config.py
@@ -161,8 +161,6 @@ CATALOG_ENABLED: bool = os.environ.get("CATALOG_ENABLED", "false").strip().lower
 CATALOG_TIER: str = (
     os.environ.get("CATALOG_TIER", "standard").strip().lower()
 )  # "standard" or "extended"
-# TTL in seconds for the extended prompt-cache tier (Anthropic API requires an integer).
-CATALOG_CACHE_TTL_SECONDS: int = int(os.environ.get("CATALOG_CACHE_TTL_SECONDS", "3600"))
 
 # Cap on how many characters the get_video_transcript tool returns to the
 # model. Long videos can produce 40K+ tokens of transcript; beyond ~30K
--- a/app/backend/rag/catalog.py
+++ b/app/backend/rag/catalog.py
@@ -9,10 +9,13 @@ from __future__ import annotations
 
 import logging
 
-from backend.config import CATALOG_CACHE_TTL_SECONDS
 from backend.db import repository
 
 logger = logging.getLogger(__name__)
+
+# Anthropic's cache_control.ttl accepts only the duration strings "5m" or
+# "1h" — not an integer count of seconds.
+EXTENDED_CACHE_TTL = "1h"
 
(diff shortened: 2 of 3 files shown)


coleam00:

An exemplary small-fix PR. It demonstrates real understanding of the Anthropic prompt-cache API contract: rather than trusting the local comment or stringifying the integer (the documented trap), it maps the tier directly to the "1h" duration enum and removes the now-incorrect 'TTL in seconds' abstraction at its source in config.py. Scope is surgical, the new constant is self-documenting, and the test suite was genuinely strengthened — the replacement regression test names the failure mode and asserts against reverting to an integer.

Weaknesses are negligible. The implementation tracks the plan with near-perfect fidelity, and the plan itself was thorough (it even noted the inert docker-compose env passthrough and correctly deferred it as deploy-scoped). The only nitpicks are pre-existing and out of scope: there is no guard that CATALOG_TIER is constrained to its two valid values, and the dead env passthrough remains for a follow-up. Neither detracts from a clean, correct, root-cause fix.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends the valid Anthropic duration string "1h"; standard tier still emits {"type": "ephemeral"} with no ttl; block content untouched. Hits every acceptance criterion exactly. |
| root_cause | 10/10 | Correctly recognizes the 'TTL in seconds' abstraction is itself wrong and removes CATALOG_CACHE_TTL_SECONDS rather than stringifying it; the tier becomes the selector mapping to the API enum. True root cause, not a band-aid. |
| scope_discipline | 10/10 | Only the three files the change requires (catalog.py, config.py, test_catalog.py); +16/-12 lines. No drive-by edits, and the dead docker-compose env passthrough was correctly left for a follow-up per deploy/ rules. |
| subtle_correctness | 9/10 | Avoids the tempting "3600" stringify trap, keeps the standard path free of a ttl key, and the regression test explicitly guards against reverting to an int. Minor: no validation that CATALOG_TIER only takes the two valid values, but that's pre-existing and out of scope. |
| code_quality | 9/10 | Clean module-level EXTENDED_CACHE_TTL constant with an accurate comment replacing the previously misleading one; idiomatic to the file's existing tier branching. The magic string is now self-documenting. |
| test_discipline | 9/10 | Updated the two integer-asserting tests and replaced the misnamed *_ttl_is_integer test with a properly-named regression test that asserts "1h" and guards against an int. Tests changed to reflect correct behavior, not gamed to pass. |
| plan_impl_fidelity | 10/10 | The diff executes the plan line-for-line: constant removed from config, EXTENDED_CACHE_TTL added, import dropped, assignment changed, and both tests updated exactly as specified. Essentially zero drift. |

---

PR #357, opened 2026-06-10

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `7ea3cbbf-40ef-4412-8b84-7bd2435bad40`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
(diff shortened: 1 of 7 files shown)


coleam00:

An exemplary fix that demonstrates genuine understanding of the Anthropic prompt-cache API contract rather than trusting the misleading local comment. It correctly identifies that the bug is not just the value but the whole "TTL in seconds" abstraction, and removes CATALOG_CACHE_TTL_SECONDS entirely while mapping the extended tier directly to the "1h" enum. It sidesteps the tempting str(3600) trap, cleans up the dead import to keep lint green, and ships a regression test that pins the exact contract and would catch both the original integer bug and the stringified-int near-miss. Implementation is a faithful, line-for-line execution of a strong plan.

The only blemish is PR-level noise: four benchmark-*.yaml harness files account for the vast majority of the diff's line count. These are benchmark scaffolding rather than implementer scope creep, so they don't reflect on the quality of the fix itself, but they do make the PR larger than the ~12-line change warrants. Net: a near-perfect surgical fix on a reasoning-heavy issue.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control.ttl = "1h" (the valid Anthropic duration enum), standard tier still emits plain {"type": "ephemeral"} with no ttl, and the catalog block text/inclusion logic is untouched. Hits every acceptance criterion exactly. |
| root_cause | 10/10 | Treats the tier as the API selector and deletes the wrong CATALOG_CACHE_TTL_SECONDS "seconds" abstraction rather than stringifying it. Correctly recognizes the config knob can never produce a valid value and removes it — the true structural fix, not a band-aid. |
| scope_discipline | 8/10 | The actual code change is surgical: catalog.py, config.py, and test_catalog.py only. The 4 new benchmark-*.yaml files (~620 of the 628 added lines) are benchmark-harness scaffolding bundled by the workflow, not implementer scope creep, but they do inflate the PR diff. |
| subtle_correctness | 9/10 | Anticipates the tempting-wrong str(3600) fix and explicitly rejects it, removes the now-unused import to avoid an F401 lint failure, and the constant removal is backward-compatible (a stale env var is simply ignored). No foot-guns introduced. |
| code_quality | 9/10 | Clean, idiomatic edit with an accurate replacement comment documenting the real API contract. Removes the misleading old comment that encoded the wrong belief. Production-grade. |
| test_discipline | 9/10 | Rewrites the regression test to pin ttl == "1h" and assert it's a str — rejecting both the original int bug and the stringified-int trap — and renames it accordingly. Also updates the equality assertion in the sibling test. Meaningful, targeted, not gamed. |
| plan_impl_fidelity | 10/10 | The diff executes source-plan.md verbatim: "1h" mapping, import removal, config constant deletion, and both test rewrites match the plan's specific changes line-for-line. Complete fidelity with zero drift. |

---

PR #358, opened 2026-06-10

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `5d3ffca8-64be-457b-b04a-b9989d13e82a`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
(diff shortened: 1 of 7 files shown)


coleam00:

A near-textbook fix for issue #246. The implementation correctly identifies that Anthropic's cache_control.ttl takes a duration enum ('5m'/'1h'), not an integer count of seconds, and maps the extended tier directly to '1h' while removing the misleading CATALOG_CACHE_TTL_SECONDS abstraction from config.py — exactly the reasoning the issue was designed to probe, and exactly what the plan prescribed (10/10 plan fidelity). Standard-tier behavior is preserved, the corrected comment documents the real contract, and the tests are updated meaningfully rather than gamed (the old int-asserting test is replaced with one asserting the duration string).

The single notable weakness is scope: the implement node's `git add -A` swept four unrelated benchmark workflow YAML files (FF/FK/FO/OF, ~619 lines) into the PR. They are benchmark-harness noise, not part of the issue, and a self-review pass should have unstaged them. The substantive code change is otherwise impeccable — surgical, correct, and well-tested. Net: an excellent fix marred only by stray harness files in the commit.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control['ttl'] = '1h' (valid Anthropic duration), standard tier keeps plain {'type': 'ephemeral'} with no ttl, and the catalog block text/inclusion is untouched. Hits every acceptance criterion exactly. |
| root_cause | 10/10 | Maps the tier to Anthropic's duration enum rather than stringifying the integer, and deletes the false 'TTL in seconds' abstraction (the CATALOG_CACHE_TTL_SECONDS constant) entirely. Avoids the tempting '3600' band-aid the issue warned about — a true structural fix. |
| scope_discipline | 5/10 | The code change itself is surgical (catalog.py + config.py + test_catalog.py, the exact 3 files needed). But the PR also carries ~619 lines of unrelated benchmark workflow YAMLs (FF/FK/FO/OF) swept in by the implement node's `git add -A`, which have nothing to do with issue #246. |
| subtle_correctness | 9/10 | Correct literal '1h', dict mutation is sound, comment now accurately states the valid enum values. Removing the import/constant leaves no dangling references. No foot-guns introduced. |
| code_quality | 9/10 | Clean and idiomatic: dead import and misleading config constant removed, the inline comment is corrected to document the actual API contract ('5m' or '1h'). Reads as production-grade. |
| test_discipline | 9/10 | Updates the happy-path assertion to '1h' and meaningfully rewrites the previously-misleading test (which encoded the wrong int contract) into one that asserts the correct Anthropic duration string, with an apt rename. Targeted and not gamed; no new edge-case test but coverage of the change is solid. |
| plan_impl_fidelity | 10/10 | The diff executes the plan line-for-line: same three files, same '1h' literal, same removal of the import and config constant, and the exact test rename/assertion changes the plan specified. Full fidelity. |

---

PR #359, opened 2026-06-10

### LINKED ISSUE #246: fix(rag): catalog extended-tier cache ttl sent as int seconds, not Anthropic duration

coleam00:
Fixes #246

**Source run-id:** `dc0da8a4-a27c-4fa3-9253-145149a14367`

### DIFF
--- a/.archon/workflows/benchmark-FF.yaml
+++ b/.archon/workflows/benchmark-FF.yaml
@@ -1,0 +1,155 @@
+name: benchmark-FF
+description: |
+  Benchmark cell: plan=Fable, impl=Fable (claude-fable-5 both slots).
+  The premium-tier ceiling. See benchmark-OO.yaml for the full design notes.
+  This file differs only in the plan and implement node provider/model
+  overrides + the cell label.
+
+provider: claude
+model: sonnet
+
+nodes:
+  - id: fetch-issue
+    bash: |
+      set -euo pipefail
+      ISSUE_NUM=$(echo "$ARGUMENTS" | grep -oE '[0-9]+' | head -1)
+      if [ -z "$ISSUE_NUM" ]; then
+        echo "ERROR: No issue number in arguments: $ARGUMENTS" >&2
+        exit 1
+      fi
+      echo "$ISSUE_NUM" > "$ARTIFACTS_DIR/.issue-number"
+      gh issue view "$ISSUE_NUM" --json number,title,body,labels,author,createdAt
+    timeout: 20000
+
+  - id: explore
+    prompt: |
+      You are a codebase explorer. Read the issue and gather context needed
+      to plan a fix. Do NOT edit anything in this node.
+
+      ## Issue
+      $fetch-issue.output
+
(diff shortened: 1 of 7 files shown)


coleam00:

A textbook-quality fix for a reasoning-trap issue. The implementer (Kimi, on a Fable plan) correctly identified that Anthropic's cache_control.ttl takes the duration enum "1h" — not an integer count of seconds and not the tempting stringified "3600" — and went a step further to recognize that the underlying CATALOG_CACHE_TTL_SECONDS abstraction was itself the bug, removing the dead constant and its misleading import/comment rather than papering over the symptom. The change is surgical (three files), the tests were genuinely inverted to lock in the correct contract (including a rename of the test that encoded the old wrong belief), and the diff faithfully tracks the plan with no drift.

The only marks against it are minor and not attributable to the fix itself: the PR diff carries four benchmark-*.yaml harness files that are orchestration scaffolding rather than issue work, which inflates the changed-file count but is expected noise from the benchmark pipeline. The plan also flagged the deploy/docker-compose.yml env-var passthrough as a worthwhile follow-up; correctly left out of scope here. Overall an excellent, root-cause, well-tested result.

## Scores

|---|---|---|
| solves_issue | 10/10 | Extended tier now sends cache_control.ttl = "1h", standard tier keeps the bare {"type": "ephemeral"} with no ttl, and the catalog block text/inclusion is untouched. Hits every acceptance criterion exactly. |
| root_cause | 10/10 | Recognized the deeper flaw — the 'TTL in seconds' config abstraction is itself wrong — and removed CATALOG_CACHE_TTL_SECONDS entirely rather than stringifying it to the still-invalid "3600". Maps the tier to the API enum, the structurally correct fix. |
| scope_discipline | 9/10 | The actual code fix is surgical: exactly config.py, catalog.py, and test_catalog.py, two lines removed from config and a one-line change in catalog. The four added benchmark-*.yaml files are benchmark-harness scaffolding, not implementer scope, so they don't reflect on the fix's discipline. |
| subtle_correctness | 9/10 | Avoided the tempting stringify trap, dropped the now-dead import so no AttributeError/dead reference lingers, and the new comment correctly documents the "5m"/"1h" contract. No foot-guns introduced. |
| code_quality | 9/10 | Clean, idiomatic edit with an accurate inline comment replacing the previously-wrong one. Reads exactly like the surrounding code. |
| test_discipline | 9/10 | Both tests were meaningfully inverted to assert the correct behavior — not gamed. The misnamed test_..._ttl_is_integer was renamed to _is_anthropic_duration, the stale patch() of the deleted symbol was removed, and an isinstance(str) check plus a documenting docstring were added. |
| plan_impl_fidelity | 10/10 | The implementation executes the Fable plan essentially line-for-line: same ttl="1h" literal and comment, same removal of the config constant and the catalog import, same test rename and assertions. No drift between plan and diff. |

---

